PR #12, opened 2017-09-18
Refactor supervisor to receive flags

stefanv:
Previously, we had separate configurations for the standard/debug
versions of each service.  Now, the $FLAGS environment variable is
sent to the services that require options instead.

Supervisor cannot handle non-existent environment variables, so
you *have* to specify FLAGS, even if it is empty.  This is addressed
in the Makefile, by modifying the supervisord script to be `FLAGS=$(FLAGS)
supervisord` instead, so that it reuses any existing FLAGS definition,
or defines it as empty otherwise.

This change will necessitate modifications to Cesium and others
relying on baselayer.

### DIFF
--- a/conf/supervisor/app.conf
+++ b/conf/supervisor/app.conf
@@ -1,8 +1,0 @@
-[include]
-files = common.conf
-
-[program:app]
-command=/usr/bin/env python baselayer/services/app.py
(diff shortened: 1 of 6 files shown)


stefanv:
I also snuck in a few PEP8 fixes to watch_logs in there.

PR #14, opened 2017-09-27
Refactor config loading into (environment + config) loading

stefanv:
- The environment (command line flags) determine which configuration
  to load.  Everywhere where the configuration is loaded, the
  environment should be examined as well.  As such, these two
  operations have been combined.

- Factoring out the config+env loader allows us to print the
  environment after launching the server, so that users know which URL
  to connect to, e.g.

- The refactor also simplifies `services/app` (which should be fairly
  minimal), and moves some of its logic to `baselayer/app/app_server`
  instead.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -1,8 +1,8 @@
 import yaml
 import os
 from pathlib import Path
-import glob
 import collections
+
 
 def recursive_update(d, u):
     # Based on https://stackoverflow.com/a/3233356/214686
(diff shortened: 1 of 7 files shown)


stefanv:
This PR will be followed by a SkyPortal and Cesium PR to make use of the environment printing.

bnaul:
I think the baselayer tests just need some environment variables to be set? Currently getting `None` for `cfg`.

Also, while we're changing this stuff, do you think we could make `load_config` error if the config files don't exist? In the current setup we see errors like the one [here](https://travis-ci.org/cesium-ml/baselayer/builds/280654820?utm_source=github_status&utm_medium=notification), `TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'`; it would be much more informative if this were just a `FileNotFound` exception.

app/config.py
```diff
@@ -60,15 +66,19 @@ def show(self):
 
 def load_config(config_files=None):
     basedir = Path(os.path.dirname(__file__))/'..'
-
     if config_files is None:
-        config_files = [Path(basedir/'config.yaml'),
-                        Path(basedir/'../config.yaml')]
+        config_files = []
```
bnaul: Why not just make `config_files=[]` the default?

PR #16, opened 2017-10-13
Add auth_or_token decorator

acrellin:


app/handlers/base.py
```diff
@@ -85,7 +86,11 @@ def push_all(self, action, payload={}):
         self.flow.push('*', action, payload=payload)
 
     def get_json(self):
-        return tornado.escape.json_decode(self.request.body)
+        try:
+            data = tornado.escape.json_decode(self.request.body)
```
bnaul: Maybe just `return` inside the `try`/`except`?

app/handlers/base.py
```diff
@@ -2,6 +2,7 @@
 import tornado.ioloop
 from sqlalchemy.orm import joinedload
 from sqlalchemy.orm.exc import NoResultFound
+import json.decoder
```
bnaul: Slightly prefer `from json.decoder import JSONDecodeError` just to clarify that we're not calling their decoder directly.

app/models.py
```diff
@@ -159,4 +159,15 @@ def is_active(self):
         return True
 
 
+class Token(Base):
+    id = sa.Column(sa.String, nullable=False, primary_key=True)
+    user_id = sa.Column(sa.ForeignKey('users.id', ondelete='CASCADE'),
+                                     nullable=False)
+    user = relationship('User', uselist=False, foreign_keys=[user_id])
```
bnaul: Is `uselist` necessary here?

app/access.py
```diff
@@ -16,3 +17,20 @@ def wrapper(self, *args, **kwargs):
...
     return check_acls
+
+
+def auth_or_token(method):
+    @functools.wraps(method)
+    def wrapper(self, *args, **kwargs):
+        data = self.get_json()
+        if 'token' in data:
+            matching_tokens = list(Token.query.filter(Token.id ==
+                                                      data['token']))
```
bnaul: `.one()` instead of `list(...)` will give you the matching item or error if there's more than one; `.first()` will just give the first one. I think either would work here.

bnaul: Actually: if this you're querying by the primary key (`id` in this case) you can just do `Token.query.get(data['token'])` as a shorthand.

PR #18, opened 2017-10-19
Parameterize ports

acrellin:


app/handlers/base.py
```diff
@@ -86,12 +86,15 @@ def push_all(self, action, payload={}):
...
 
     def get_json(self):
-        try:
-            return tornado.escape.json_decode(self.request.body)
-        except JSONDecodeError:
-            print(' ** get_json failed on request with JSONDecodeError - '
-                  'returning {} ** ')
+        if len(self.request.body) == 0:
             return {}
+        else:
```
bnaul: No need for this `else`, dedent this block

app/flow.py
```diff
@@ -1,12 +1,16 @@
 import zmq
 from .json_util import to_json
+from .env import load_env
+
+
+env, cfg = load_env()
```
bnaul: @stefanv can you think of any issues w/ loading the config at import time like this?

app/models.py
```diff
@@ -160,7 +161,8 @@ def is_active(self):
 
 
 class Token(Base):
-    id = sa.Column(sa.String, nullable=False, primary_key=True)
+    id = sa.Column(sa.String, default=str(uuid.uuid4()), nullable=False,
```
bnaul: This should be `default=lambda: str(uuid.uuid4())`; right now you'll get the same UUID every time.

app/flow.py
stefanv: Should be fine, we no longer rely on magical things happening in the background.

acrellin:
(@stefanv's commit)

stefanv:
Haha, perfect.  I read through this change several times, thinking Ari will *not* be impressed if I break his code.  Well...

stefanv:
OK, and then do `git push -f acrellin master` to complicate things, when your local branch is named `acrellin-master` ;)

stefanv:
And then learn you cannot re-open the PR by pushing, unless you are Ari.  Grr.

```
To github.com:acrellin/baselayer.git
 ! [remote rejected] acrellin-master -> master (permission denied)
error: failed to push some refs to 'git@github.com:acrellin/baselayer.git'
```

acrellin:
lol, I'll sort it out in the morning 🙃

acrellin:
no worries @stefanv 😉

PR #22, opened 2017-10-24
Parameterize ports (v2)

stefanv:


stefanv:
Reopening so that we can add the automated config file template hooks.

PR #23, opened 2017-10-24
Centralize Makefile logic in baselayer

bnaul:


acrellin:
Also nice

tools/fill_conf_values.py
```diff
@@ -2,13 +2,16 @@
...
 from baselayer.app.env import load_env
 
 
-def fill_config_file_values(
-        template_paths=['baselayer/conf/nginx.conf.template',
-                        'baselayer/conf/supervisor/supervisor.conf.template']):
+def fill_config_file_values(template_paths=None):
+    if template_paths is None:
+        basedir = pathlib.Path(os.path.dirname(__file__))/'..'
+        template_paths = [basedir/'conf/nginx.conf.template',
```
stefanv: Shall we make this operate on an explicitly specified filename, as discussed earlier?

stefanv:
MIssing `test_frontend.py`

tools/test_frontend.py
```diff
@@ -0,0 +1,91 @@
...
+               (not all([b'RUNNING' in line for line in supervisor_status()]))):
+            time.sleep(1)
+            timeout += 1
+
+        if timeout == 10:
+            print('[test_frontend] Could not launch server processes; terminating')
+            sys.exit(-1)
+
+        for timeout in range(10):
+            conn = http.HTTPConnection("localhost", 5000)
```
stefanv: This should grab the port from the config file.

PR #25, opened 2017-10-25
Simplify server availability checking

bnaul:


tools/test_frontend.py
```diff
@@ -28,6 +25,30 @@
 TEST_CONFIG = 'test_config.yaml'
 
 
+def all_services_running():
```
stefanv: Let's also pop in a docstring or note here.  Otherwise, it looks quite odd that we are checking whether services either are running or exited, and then report success.  It may not be obvious to a reader that there are other exit statuses.

tools/test_frontend.py
```diff
@@ -28,6 +25,30 @@
...
+            assert all_services_running(), ("Webservice(s) failed to launch:\n"
+                                            + '\n'.join(supervisor_status()))
+            response = requests.get(url)
+            assert response.status_code == 200, ("Expected status 200, got" 
+                                                 f" {response.status_code}"
+                                                 f" for URL {url}.")
+            return  # all checks passed
+        except Exception as e:
+            if i == max(range(timeout)):  # last iteration
+                raise ConnectionError(str(e)) from None
```
stefanv: Should we include URL?

PR #26, opened 2017-10-25
Launch WebPack as required by server mode

### LINKED ISSUE #24: Webpack microservice should execute webpack once for non debug run

stefanv:
1. If in production mode, launch webpack only if `bundle.js` is missing.
2. When in debug mode, launch the webpack monitor (changes to .js
   files are automatically re-bundled).

Closes #24

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,13 +57,16 @@ run: paths dependencies fill_conf_values
 	echo && \
 	$(SUPERVISORD)
 
-run_production: paths dependencies fill_conf_values bundle
+run_production: paths fill_conf_values
+	@echo
+	@echo "[!] Production run: not automatically installing dependencies."
+	@echo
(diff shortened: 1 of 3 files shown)


bnaul:
So the idea here is that, to run in production, you'd manually call `make dependencies` and `make bundle`, then (optionally) remove `node_modules` and call `make run_production`? That's basically the series of steps in the `Dockerfile`.

stefanv:
Right, so `make run_production` is for expert users who have everything set up and ready to go.  I can make the pre-required steps clearer, if that would help; if so, where should that be documented?

bnaul:
Skyportal docs would be most important I guess? I think that can just go along with whatever documentation you end up writing for general deployment stuff, doesn't need to be part of this PR.

PR #51, opened 2019-03-12
Try serving files statically via Nginx before reaching out to backend Python server

stefanv:
Prefer static serving of files through NGINX
    
    NGINX will now check for the existence of a static file, before
    requesting a page from the backend server.
    
    Also made small tweaks to the config so that the root URL is specified
    similar to the websocket and auth endpoints.


stefanv:
Confirmed that this is an actual issue with this PR.  Will have to investigate a bit more deeply.

PR #53, opened 2019-03-14
Obtain auth token from Authorization header, instead of payload

stefanv:
Once this is merged, I will submit a test to skyportal to verify its behavior.


acrellin:
I believe skyportal will need to be updated in several places to use this approach

PR #56, opened 2019-05-22
Config loading: allow missing files; remove unused quiet flag

stefanv:

PR #57, opened 2019-05-31
Warn about using default configs, allow missing configs

denniscfeng:
Continuation of https://github.com/cesium-ml/baselayer/pull/56/files to address https://github.com/skyportal/skyportal/issues/69, warning about usage of default configs.


stefanv:
:+1:

PR #61, opened 2019-07-03
Enable batch mode on DB

dannygoldstein:
Psycopg2 Batch Mode (Fast Execution)
Modern versions of psycopg2 include a feature known as Fast Execution Helpers , which have been shown in benchmarking to improve psycopg2’s executemany() performance with INSERTS by multiple orders of magnitude. SQLAlchemy allows this extension to be used for all executemany() style calls invoked by an Engine when used with multiple parameter sets, by adding the use_batch_mode flag to create_engine():

engine = create_engine(
    "postgresql+psycopg2://scott:tiger@host/dbname",
    use_batch_mode=True)
Batch mode is considered to be experimental at this time, however may be enabled by default in a future release.

See also

Executing Multiple Statements - demonstrates how to use DBAPI executemany() with the Connection object.

New in version 1.2.0.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -20,7 +20,7 @@ class Config(dict):
         dict.__init__(self)
         if config_files is not None:
             cwd = os.getcwd()
-            config_names = [Path(c).relative_to(cwd) for c in config_files]
+            config_names = [Path(c) for c in config_files]
             print(f'  Config files: {config_names[0]}')
             for f in config_names[1:]:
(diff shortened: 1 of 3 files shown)


dannygoldstein:
https://docs.sqlalchemy.org/en/13/dialects/postgresql.html#psycopg2-batch-mode

dannygoldstein:
dead PR

PR #64, opened 2019-10-01
Add push_notification method to base handler

acrellin:


acrellin:
Will update skyportal accordingly when this is merged.

stefanv:
Would be nice to add a test for this in Baselayer.  Otherwise :+1: .

acrellin:
We don't currently have any testing set up for baselayer, do we?  I can add
a corresponding test to skyportal.

stefanv:
I think we do. It's supposed to deploy the test application.

PR #65, opened 2019-10-04
Allow specification of config files for db_init via FLAGS

stefanv:


acrellin:
This looks great. Did you also want to add this to `db_init` in addition to `db_clear`?

Makefile
```diff
@@ -38,7 +38,7 @@ db_init: dependencies
 
 db_clear: ## Delete all data from the database.
 db_clear: dependencies
```
acrellin: Should this also be added to `db_init`?

stefanv:
Good point; done.

PR #66, opened 2019-10-10
Set FLAGS to config.yaml for all targets

stefanv:
This allows FLAGS to be extended from the outside, so you can do, e.g.:

```
FLAGS="--config=foo.yaml" make run
```

We never fully supported this use case before, relying on `config.yaml` to be the default configuration.


stefanv:
That's a good point, Ari.  I've updated the PR to use the logic discussed.

PR #67, opened 2019-10-10
Allow disabling dask through configuration

stefanv:
Closes https://github.com/skyportal/skyportal/issues/245 and potentially https://github.com/skyportal/skyportal/issues/164


acrellin:
This looks great!

PR #75, opened 2019-11-02
Update API Logging

jialin-wu-02:
When API was called, a message detailing the uri of the API, the name of the user, and the time of the API call would be logged.


acrellin:
@jialin-wu-02 As we discussed, this should apply to `permissions` as well.

acrellin:
@jialin-wu-02 Should all of these commits be included in this PR? Can you try rebasing or cherry-picking the relevant commits onto a fresh master-based branch?

PR #77, opened 2020-01-29
Use pip-installed supervisor

acrellin:


stefanv:
Only the one small change we discussed this morning, then I think we're good to go.

acrellin:
That wasn't clear - can you please explain (or provide an example in a
comment)?

acrellin:
(It's unclear how a format string would add clarity)

tools/supervisor_status.py
```diff
@@ -10,7 +10,7 @@
 
 def supervisor_status():
     result = subprocess.run(
-        'supervisorctl -c baselayer/conf/supervisor/supervisor.conf status',
+        'python -m supervisor.supervisorctl -c baselayer/conf/supervisor/supervisor.conf status',
         shell=True, cwd=base_dir, check=True,
```
stefanv: Maybe something like:

```
# top of the file
config = 'baselayer/conf/supervisor/supervisor.conf'
supervisorctl = ['python', '-m', 'supervisor.supervisorctl', '-c', config]

# further down
result = subprocess.run(supervisorctl + ['status'], ...)
```

acrellin:
Even though we never run this command with any other options? This extra
verbosity makes it less clear to me, actually.

acrellin:
If it were something we were doing multiple times with different options in
this file, then I could see the case for that.

PR #85, opened 2020-03-05
Have get_if_owned_by return None if no obj w/ ID instead of erroring

acrellin:
Previously, if an ID was supplied to `Base.get_if_owned_by` that did not exist, an error was raised. This method should mirror the behavior of `query.get`, which returns `None` instead of erroring in that case. This PR makes `get_if_owned_by` return `None` if no matching ID exists, to match the behavior of `query.get`.


stefanv:
Fine with this change, just check that baselayer_template_app and skyportal both still do their thing!

PR #86, opened 2020-03-17
Add anacron-style job execution

stefanv:


acrellin:
@stefanv this looks great - do we need to add this service to the supervisord config, too?

stefanv:
Err, yes!! Good catch.

acrellin:
@stefanv how difficult would it be to also specify the time of day the job should be run?  E.g., expensive jobs might best be run during non-peak hours

stefanv:
@acrellin Yes, we can add functionality such as "should be run every x hours, but only between time a and time b".

PR #88, opened 2020-04-06
Update Logging API Calls

jialin-wu-02:
Log the method name, the time of the call, and the username of any APIs that the user calls.

### DIFF
--- a/app/access.py
+++ b/app/access.py
@@ -26,7 +26,8 @@ def auth_or_token(method):
         if cfg["server"]["log"]["api_calls"]:
             log = structlog.get_logger()
             log.msg("Api Called",
-                    created_at=(self.current_user.created_at).strftime('%m/%d/%Y %H:%M:%S'),
+                    created_at=(self.current_user.created_at)
+                    .strftime('%m/%d/%Y %H:%M:%S'),
                     api_uri=self.request.uri,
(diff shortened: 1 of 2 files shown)


dannygoldstein:
When merged and reconciled against master, this breaks the skyportal test suite as `self.current_user` can be undefined for some endpoints (such as . There is a simpler way to do this by modifying the `nginx` configuration file. Will submit a PR. 

```
ERROR:tornado.application:Uncaught exception GET /api/filters/1 (127.0.0.1)
HTTPServerRequest(protocol='http', host='localhost:5000', method='GET', uri='/api/filters/1', version='HTTP/1.0', remote_ip='127.0.0.1')
Traceback (most recent call last):
  File "/anaconda2/envs/skyportal2/lib/python3.7/site-packages/tornado/web.py", line 1697, in _execute
    result = method(*self.path_args, **self.path_kwargs)
  File "/Users/danny/PycharmProjects/skyportal/baselayer/app/access.py", line 29, in wrapper
    created_at=(self.current_user.created_at)
AttributeError: 'NoneType' object has no attribute 'created_at'
```

stefanv:
@dannygoldstein I still think it's useful to have the username in the log, when available? We know more about the request here than what nginx does.

PR #105, opened 2020-06-24
Separate services and allow enabling/disabling

stefanv:
Please see [the documentation update](https://github.com/cesium-ml/baselayer/blob/daa16f973e00147882a06247fbc7fd0e67106446/doc/usage.md#microservices) for more information.

@dmitryduev This should allow for easier customization of microservices during deployment, so that we can, e.g., run dask on one pod and the web server on another.  

@dannygoldstein This was done with the follow up service in mind, e.g..


stefanv:
@acrellin Any thoughts on this PR?

tools/setup_services.py
```diff
@@ -0,0 +1,60 @@
...
+
+    both = set().union(disabled).intersection(enabled)
+    if both:
+        raise RuntimeError(
+            f'Invalid service specification: {both} in both enabled and disabled'
+        )
+
+    if disabled == '*':
+        disabled = services.keys()
+    if enabled == '*':
```
acrellin, line 40: Why does the enabled list get set as empty if all services are set to be enabled?

stefanv: The logic is a few lines down. It wouldn't hurt to set it to the full list, it's just that we start with the full list (so, that's the default), take away as necessary from disabled, then add back any specified services.

Since the default is all services, setting enabled to star is equivalent to not modifying the default.

acrellin:
This is great, thanks @stefanv!

PR #118, opened 2020-07-31
Update User model

profjsb:
This PR adds basic information, like name, contact email, etc. to the User model. This will be heavily used upcoming Skyportal PRs.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1,5 +1,6 @@
 from datetime import datetime
 import uuid
+from hashlib import md5
 
 import sqlalchemy as sa
 from sqlalchemy.ext.associationproxy import association_proxy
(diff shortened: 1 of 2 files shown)


app/access.py
```diff
@@ -40,7 +40,8 @@ def check_acls(method):
         @auth_or_token
         @functools.wraps(method)
         def wrapper(self, *args, **kwargs):
-            if not set(acl_list).issubset(self.current_user.permissions):
+            if not (set(acl_list).issubset(self.current_user.permissions)
+                    or "System admin" in self.current_user.permissions):
```
acrellin, line 44: Looks like this needs a rebase/merge

profjsb: Frustrating--I did a rebase. What am I doing wrong?

PR #119, opened 2020-08-01
Include route in error messages

stefanv:
This should make it possible to track down where errors occur


profjsb:
Good idea.

acrellin:
This will be very helpful, thanks

PR #120, opened 2020-08-04
watch app log in watch_logs.py to aid debugging, especially in travis

dannygoldstein:


dannygoldstein:
Causes travis builds to be too messy

PR #122, opened 2020-08-09
Make dbinit more robust and provide helpful feedback

stefanv:
- Before, we did not fail when db already existed.  Now, there is an
  explicit check for that.  We also don't try and recreate an existing
  database.
- Explain to the user how to create a DB and permissions if our
  creation didn't work.  System configurations can be complex, so we
  cannot expect to succeed always.

### DIFF
--- a/tools/db_init.py
+++ b/tools/db_init.py
@@ -39,50 +39,18 @@ if host:
 if port:
     flags += f' -p {port}'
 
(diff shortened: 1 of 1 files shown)


acrellin:
@stefanv is this good to go?

stefanv:
Yeap! Can you try it locally first?

stefanv:
@acrellin Rebased.  Please try locally first.

PR #127, opened 2020-08-11
Wait for WebSocket before proceeding with tests

stefanv:


app/test_util.py
```diff
@@ -38,6 +37,12 @@ def server_url(self, value):
 
     def get(self, uri):
         return webdriver.Firefox.get(self, self.server_url + uri)
+        webdriver.Firefox.get(self, self.server_url + uri)
```
acrellin, line 40: How can this ever be executed if it's right after a return statement?

stefanv: Yeah, that return is supposed to go.  Didn't grab that line of the patch, will do.

stefanv:
@ari Fixed the merge conflict.

ari:
Who are you?

stefanv:
We meet the real @ari! :) 

Sorry for that, @ari, I was trying to tag teammate Ari who is @acrellin.

PR #128, opened 2020-08-11
Slight tweaks on app utils

stefanv:


app/test_util.py
```diff
@@ -50,13 +50,13 @@ def wait_for_css(self, css, timeout=5):
...
-        return WebDriverWait(self, timeout).until_not(
-            expected_conditions.presence_of_element_located((By.XPATH, xpath))
+        return WebDriverWait(self, timeout).until(
+            expected_conditions.invisibility_of_element((By.XPATH, xpath))
         )
 
     def wait_for_css_to_disappear(self, css, timeout=5):
         return WebDriverWait(self, timeout).until_not(
-            expected_conditions.presence_of_element_located((By.CSS, css))
+            expected_conditions.invisibility_of_element((By.CSS, css))
```
acrellin, line 59: This is `until_not` and `invisibility` -- should it be `until` above?

app/test_util.py
```diff
@@ -69,13 +69,13 @@ def wait_for_css_to_be_clickable(self, css, timeout=5):
             expected_conditions.element_to_be_clickable((By.CSS, css))
         )
 
-    def scroll_to_element_and_click(self, element):
-        self.execute_script("arguments[0].scrollIntoView();", element)
-        return element.click()
+    def scroll_to_element_and_click(self, element, timeout=5):
```
acrellin, line 72: Where is the timeout being used?

app/test_util.py
```diff
@@ -69,13 +69,13 @@ def wait_for_css_to_be_clickable(self, css, timeout=5):
...
+        self.execute_script("arguments[0].scrollIntoView(true);", element)
+        ActionChains(self).move_to_element(element).click().perform()
 
     def click_xpath(self, xpath):
         element = self.wait_for_xpath_to_be_clickable(xpath)
-        return self.scroll_to_element_and_click(element)
+        self.scroll_to_element_and_click(element)
```
acrellin, line 78: Shouldn't this return whatever click returns?

app/test_util.py
stefanv: Correct!

app/test_util.py
stefanv: We don't use it, but probably good practice.

PR #131, opened 2020-08-13
Bump timeout defaults to 10 s

acrellin:


acrellin:
Merging as we discussed and okayed this earlier

PR #135, opened 2020-08-19
Document models

dannygoldstein:
This PR documents each class in `baselayer.app.models`. The methods, relationships, properties, and `Column`s of each documented class are also documented.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -48,12 +48,24 @@ def init_db(user, database, password=None, host=None, port=None):
 
 class BaseMixin(object):
     query = DBSession.query_property()
-    id = sa.Column(sa.Integer, primary_key=True)
-    created_at = sa.Column(sa.DateTime, nullable=False, default=utcnow)
-    modified = sa.Column(sa.DateTime, default=utcnow, onupdate=utcnow, nullable=False)
+    id = sa.Column(sa.Integer, primary_key=True, doc='Unique object identifier.')
+    created_at = sa.Column(
+        sa.DateTime,
+        nullable=False,
+        default=utcnow,
+        doc="UTC time of insertion of object's row into the database.",
+    )
+    modified = sa.Column(
+        sa.DateTime,
+        default=utcnow,
+        onupdate=utcnow,
+        nullable=False,
+        doc="UTC time the object's row was last modified in the database.",
+    )
 
     @declared_attr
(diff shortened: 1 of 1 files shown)


acrellin:
@dannygoldstein did you see this? https://travis-ci.org/github/cesium-ml/baselayer/builds/719200884#L780

app/models.py
```diff
@@ -182,85 +231,192 @@ def join_model(
...
-        sa.String, nullable=False, unique=True, default=lambda: str(uuid.uuid4())
+        sa.String,
+        nullable=False,
+        unique=True,
+        default=lambda: str(uuid.uuid4()),
+        doc="The name of the token.",
     )
 
     def is_owned_by(self, user_or_token):
+        """Return a boolean indicating whether a User (or Token) has read access
```
acrellin, line 403: This checks whether a token instance is owned by the specified user (or has the same identity as the specified token, if a token is passed in here)

app/models.py
```diff
@@ -82,10 +117,24 @@ def get_if_owned_by(cls, ident, user, options=[]):
         return obj
 
     def is_owned_by(self, user):
+        """Return a boolean indicating whether a User has read access to this
+        object.
+
+        Parameters
+        ----------
+        user : `baselayer.app.models.User`
```
acrellin, line 125: `user` can also be a `Token` instance, I believe

app/models.py
```diff
@@ -68,12 +82,33 @@ def __repr__(self):
...
     def get_if_owned_by(cls, ident, user, options=[]):
+        """Return an object from the database if the requesting user has access
+        to read the object. If the requesting user does not have access, raise
+        an AccessError.
+
+        Parameters
+        ----------
+        ident : integer or string
+           Primary key of the requested object.
+        user : `baselayer.app.models.User`
```
acrellin, line 102: or `Token`

app/models.py
dannygoldstein: fixed

app/models.py
dannygoldstein: fixed

app/models.py
dannygoldstein: fixed

dannygoldstein:
@profjsb  need to merge this to merge skyportal/skyportal#729

PR #141, opened 2020-09-10
Always rebuild JS bundle upon launch

stefanv:
This doesn't fix any bug (everything was working as expected, as far as I can tell—no rebuilds on production runs); but it should help @dima in that now the bundle will rebuild (in production mode) upon initial app launch via `make run_production`.


dmitryduev:
Looks great, can we merge it now?

PR #142, opened 2020-09-10
Ensure OAuth UID matches username

stefanv:
Currently, if a server gets re-deployed with a new database, previously
logged in users are able to access the system.  Often, their UIDs now
point to a new user, so they "log in" as this new person.

To address this, we now, in addition to user_id, set a cookie for OAuth
UID (Google email). When logging in, we check this cookie to
ensure that the user doesn't get logged into a different username.

The accompanying SkyPortal patch is https://github.com/skyportal/skyportal/pull/887

### DIFF
--- a/app/handlers/base.py
+++ b/app/handlers/base.py
@@ -46,13 +46,17 @@ class PSABaseHandler(RequestHandler):
         if user_id and oauth_uid:
             user = User.query.get(int(user_id))
             sa = user.social_auth.first()
+            if sa is None:
+                # No SocialAuth entry; probably machine generated user
+                return user
             if (sa.uid.encode('utf-8') == oauth_uid):
                 return user
 
     def login_user(self, user):
-        sa = user.social_auth.first()
         self.set_secure_cookie('user_id', str(user.id))
-        self.set_secure_cookie('user_oauth_uid', sa.uid)
+        sa = user.social_auth.first()
+        if sa is not None:
+            self.set_secure_cookie('user_oauth_uid', sa.uid)
 
     def write_error(self, status_code, exc_info=None):
         if exc_info is not None:
--- a/app/handlers/base.py
+++ b/app/handlers/base.py
@@ -1,6 +1,7 @@
 import time
+import inspect
 import tornado.escape
-from sqlalchemy.orm import joinedload
+from tornado.web import RequestHandler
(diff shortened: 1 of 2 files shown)


app/handlers/base.py
```diff
@@ -20,6 +24,51 @@
...
+class PSABaseHandler(RequestHandler):
+    """
+    Mixin used by Python Social Auth
+    """
+    def user_id(self):
+        return self.get_secure_cookie('user_id')
+
+    def get_current_user(self):
+        user_id = self.user_id()
+        oauth_uid = self.get_secure_cookie('user_oauth_uid')
```
acrellin, line 45: Why define a method for `user_id` but not `user_oauth_uid`?

app/handlers/base.py
```diff
@@ -20,6 +24,51 @@
...
+        if user_id and oauth_uid:
+            user = User.query.get(int(user_id))
+            sa = user.social_auth.first()
+            if (sa.uid.encode('utf-8') == oauth_uid):
+                return user
+
+    def login_user(self, user):
+        sa = user.social_auth.first()
```
acrellin, line 53: What if there's no match? Where do new users get created?

app/handlers/base.py
```diff
@@ -20,6 +24,51 @@
...
+        else:
+            err = "An unknown error occurred"
+        self.render("loginerror.html", app=cfg["app"], error_message=str(err))
+
+
+# Monkey-patch in each method of social_tornado.handlers.BaseHandler
+for (name, fn) in inspect.getmembers(
+        PSABaseHandler, predicate=inspect.isfunction
+):
+    setattr(psa_handlers.BaseHandler, name, fn)
```
acrellin, line 69: Slick! :sunglasses:

acrellin (commented):
Looks good to me, just a couple quick questions below

app/handlers/base.py
stefanv: That method is defined in PSA, so I just kept it as is since I wasn't sure if they used it anywhere.

Do you want me to remove it?

app/handlers/base.py
stefanv: If there's no match, then the user does not get logged in, and then the pipeline proceeds the same way it usually does.

acrellin (approved):

PR #147, opened 2020-09-16
Serialize PhoneNumber types from SQLAlchemy utils

kmshin1397:
Add JSON serialization support for phone numbers from sqlalchemy_utils phone numbers

### DIFF
--- a/app/json_util.py
+++ b/app/json_util.py
@@ -42,6 +42,9 @@ class Encoder(json.JSONEncoder):
             o.columns = o.columns.droplevel('channel')  # flatten MultiIndex
             return o.to_dict(orient='index')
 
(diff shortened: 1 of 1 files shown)


acrellin:
@kmshin1397 I'm seeing that we're actually already using sqlalchemy_utils in models.py even though it's not in requirements.txt (yikes!) so we can just go ahead and add it to requirements.txt and you can import and then update check here to use `isinstance`

kmshin1397:
@acrellin I see the sqlalchemy_utils in the requirements already [here](https://github.com/cesium-ml/baselayer/blob/ada8ba555ddc88ad049c935dbe1db655fdd2a886/requirements.txt#L13) actually, so I just switched to the importing PhoneNumber. Just didn't do it initially because I was mindlessly following the two cases above it.

PR #149, opened 2020-09-24
Add external logging as a supervisord microservice

profjsb:
This PR adds the capability of streaming a subset of logs to a 3rd party provider using `SysLogHandler` from the `logging` module of Python. Currently we use [papertrail](https://www.papertrail.com) which provides a free tier of 1-week log retention and alerting integration. External logging is established as a microservice. The (secret) parameters are set up in `config.yaml`:

```yaml
external_logging:
    enabled: False
    services:
        papertrail:
           # get an account at https://papertrailapp.com
           enabled: False
           # change url to the correct subdomain for your account
           # and change the port as well
           url: logX.papertrailapp.com
           port: XX
           # which log files, if any, do you not want to send over to the 3rd party?
           excluded_log_files: []
```

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -60,6 +60,18 @@ ports:
     websocket_path_in: 'ipc://run/message_flow_in'
     websocket_path_out: 'ipc://run/message_flow_out'
 
+
+external_logging:
+    enabled: False
+    # get an account at https://papertrailapp.com
+    service_name: papertrail
+    # change url to the correct subdomain for your account
+    # and change the port as well
+    url: logX.papertrailapp.com
+    port: XX
+    # which log files, if any do you not want to send over to the 3rd party?
+    excluded_log_files: []
+
 # You can schedule jobs to run at a certain time interval (given in minutes).
 #
 # If baselayer is not running at the time the job is supposed to run,
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -63,15 +63,16 @@ ports:
 
 external_logging:
     enabled: False
-    # get an account at https://papertrailapp.com
     services:
-        - papertrail:
-            # change url to the correct subdomain for your account
-            # and change the port as well
-            url: logX.papertrailapp.com
-            port: XX
-            # which log files, if any do you not want to send over to the 3rd party?
-            excluded_log_files: []
+        papertrail:
+           # get an account at https://papertrailapp.com
+           enabled: False
+           # change url to the correct subdomain for your account
+           # and change the port as well
+           url: logX.papertrailapp.com
+           port: XX
+           # which log files, if any do you not want to send over to the 3rd party?
+           excluded_log_files: []
 
 # You can schedule jobs to run at a certain time interval (given in minutes).
 #
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -64,13 +64,14 @@ ports:
 external_logging:
     enabled: False
     # get an account at https://papertrailapp.com
-    service_name: papertrail
-    # change url to the correct subdomain for your account
-    # and change the port as well
-    url: logX.papertrailapp.com
-    port: XX
-    # which log files, if any do you not want to send over to the 3rd party?
-    excluded_log_files: []
+    services:
+        - papertrail:
+            # change url to the correct subdomain for your account
+            # and change the port as well
+            url: logX.papertrailapp.com
+            port: XX
+            # which log files, if any do you not want to send over to the 3rd party?
+            excluded_log_files: []
 
 # You can schedule jobs to run at a certain time interval (given in minutes).
 #
--- a/services/external_logging/supervisor.conf
+++ b/services/external_logging/supervisor.conf
@@ -1,7 +1,5 @@
-[program:cron]
-command=/usr/bin/env python baselayer/services/cron/cron.py %(ENV_FLAGS)s
+[program:external_logging]
+command=/usr/bin/env python baselayer/services/external_logging/external_logging.py %(ENV_FLAGS)s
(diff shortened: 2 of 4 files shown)


config.yaml.defaults
```diff
@@ -60,6 +60,18 @@ ports:
...
+external_logging:
+    enabled: False
+    # get an account at https://papertrailapp.com
+    service_name: papertrail
+    # change url to the correct subdomain for your account
+    # and change the port as well
+    url: logX.papertrailapp.com
+    port: XX
+    # which log files, if any do you not want to send over to the 3rd party?
+    excluded_log_files: []
```
dannygoldstein, line 73:
```suggestion
    services:
        - papertrail:
            # change url to the correct subdomain for your account
            # and change the port as well
            url: logX.papertrailapp.com
            port: XX
            # which log files, if any do you not want to send over to the 3rd party?
            excluded_log_files: []
```

dannygoldstein: Suggesting the above because: 

1. Other services may have different config parameters beyond url, port, excluded log files
2. We may want to support logging to multiple services

tools/watch_logs.py
```diff
@@ -10,6 +10,44 @@
...
+        if cfg['external_logging']['service_name'] != 'papertrail':
+            # we do not know how to do this yet
+            print(
+                "Warning: we only know how to use papertrail"
+                " for external logging"
+            )
+            external_logging_enabled = False
+        try:
+            if cfg['external_logging']["url"].find("papertrailapp.com") == -1:
+                print("Warning: incorrect URL for papertrail logging")
```
dannygoldstein, line 36: Why not use baselayer.log (as in other PRs) here and below?

tools/watch_logs.py
```diff
@@ -10,6 +10,44 @@
...
+from logging.handlers import SysLogHandler
+
+from app.env import load_env
+
+env, cfg = load_env()
+
+
+# check 3rd party logging, if enabled, and make sure that
+# it is set up properly
+if cfg.get('external_logging'):
```
dannygoldstein, line 24: Suggestion: validate the `external_logging` attribute using JSONschema.

profjsb: I'd rather hold off on this until we have multiple services and understand what a common schema could look like. Ok?

tools/watch_logs.py
profjsb: Done.

config.yaml.defaults
profjsb: Done.

dannygoldstein (approved):

config.yaml.defaults
```diff
@@ -60,6 +60,20 @@ ports:
     websocket_path_in: 'ipc://run/message_flow_in'
     websocket_path_out: 'ipc://run/message_flow_out'
 
+
+external_logging:
+    enabled: False
+    services:
+        papertrail:
```
stefanv, line 67: Do we need both global enable and service-specific enable?

tools/watch_logs.py
```diff
@@ -9,7 +9,59 @@
...
+
+    if not cfg.get('external_logging'):
+        return external_logging_enabled, enabled_services
+
+    external_logging_enabled = cfg['external_logging']["enabled"]
+    if not external_logging_enabled:
+        return external_logging_enabled, enabled_services
+
+    print(cfg['external_logging'])
+    for service, config in cfg['external_logging']["services"].items():
```
stefanv, line 41: Should this become a separate microservice rather?  We cannot guarantee that watch_logs.py will be run otherwise.

That allows us to also use the existing enable/disable microservice configuration.

config.yaml.defaults
profjsb: it's a convenience move if we have many services. For now it's superfluous.

tools/watch_logs.py
profjsb: Interesting idea. @dannygoldstein what do you think?

tools/watch_logs.py
```diff
@@ -82,13 +82,14 @@ def print_col(line):
...
-for t in threads:
-    t.join()
+if __name__ == "__main__":
+    colors = ['default', 'green', 'yellow', 'blue', 'magenta', 'cyan', 'red']
+    threads = [
+        threading.Thread(target=print_log, args=(logfile, colors[n % len(colors)]))
+        for (n, logfile) in enumerate(watched)
+    ]
+
+    for t in threads:
```
profjsb, line 92: No real change to this code other than to make this `watch_logs.py` importable without also kicking off the in-line log tailing.

tools/watch_logs.py
dannygoldstein: sounds good to me

dannygoldstein (approved):

services/external_logging/supervisor.conf
```diff
@@ -0,0 +1,7 @@
+[program:cron]
+command=/usr/bin/env python baselayer/services/cron/cron.py %(ENV_FLAGS)s
+environment=PYTHONPATH=".",PYTHONUNBUFFERED="1"
+startretries=0
+startsecs=1
+stdout_logfile=log/cron.log
```
acrellin, line 6: These need to be updated as they appear to be copied & unchanged from the cron service (program name, command, stdout_logfile)

acrellin (commented):
@profjsb after the required changes to supervisor.conf, :+1: from me

services/external_logging/supervisor.conf
profjsb: Good catch. Strange -- I had the correct file in there in my repo. Must have checked in the wrong version. Fixed now.

PR #151, opened 2020-10-01
Papertrail update

stefanv:


config.yaml.defaults
```diff
@@ -62,17 +62,15 @@ ports:
...
-           port: XX
-           # which log files, if any do you not want to send over to the 3rd party?
-           excluded_log_files: []
+    papertrail:
+       # get an account at https://papertrailapp.com
+       enabled: False
+       # change url to the correct subdomain for your account
+       # and change the port as well
+       url: logX.papertrailapp.com
+       port: XX
```
profjsb, line 71:
```suggestion
       port:
```

config.yaml.defaults
```diff
@@ -62,17 +62,15 @@ ports:
...
-           url: logX.papertrailapp.com
-           port: XX
-           # which log files, if any do you not want to send over to the 3rd party?
-           excluded_log_files: []
+    papertrail:
+       # get an account at https://papertrailapp.com
+       enabled: False
+       # change url to the correct subdomain for your account
+       # and change the port as well
+       url: logX.papertrailapp.com
```
profjsb, line 70:
```suggestion
       url:
```

profjsb (approved):
Looks good!

tools/watch_logs.py
```diff
@@ -45,7 +45,7 @@ def logs_from_config(supervisor_conf):
 watched.append('log/nginx-bad-access.log')
 watched.append('log/nginx-error.log')
 watched.append('log/fake_oauth2.log')
-
+watched.append('log/app_00.log')
```
profjsb, line 48: Do you want to watch other files too?

stefanv: I think there's a bigger problem we need to address here, which is that we should pick up on new logfiles as they appear and add threads for them.  I'll fix that in another PR.

PR #153, opened 2020-10-06
Migration manager

stefanv:
The migration manager checks to see whether migrations are present.
If so, it attempts to perform a migration if the DB is not up to date
already.  It then runs an HTTP server to report the current migration
status.

This status can be queried by the app factory to decide whether to
launch or not.

See also https://github.com/skyportal/skyportal/pull/1055

### DIFF
--- a/services/status_server/status_server.py
+++ b/services/status_server/status_server.py
@@ -1,0 +1,22 @@
+import tornado.ioloop
+import tornado.web
+
+from baselayer.app.env import load_env
+from baselayer.log import make_log
+
+env, cfg = load_env()
+
+
+class MainHandler(tornado.web.RequestHandler):
+    def get(self):
+        self.write("<h2>SkyPortal is being provisioned</h2>")
+
+def make_app():
+    return tornado.web.Application([
+        (r"/", MainHandler),
+    ])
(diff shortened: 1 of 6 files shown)


services/status_server/status_server.py
```diff
@@ -0,0 +1,22 @@
...
+
+from baselayer.app.env import load_env
+from baselayer.log import make_log
+
+env, cfg = load_env()
+
+
+class MainHandler(tornado.web.RequestHandler):
+    def get(self):
+        self.write("<h2>SkyPortal is being provisioned</h2>")
```
acrellin, line 12: So this always shows the same thing?

acrellin (approved):
Looks good. See my comment below about the status server always returning the same thing ("SP is being provisioned")

services/status_server/status_server.py
stefanv: Yes, it's just a backup so that we don't see the 5xx errors.  Is there something else that would be useful to show?

stefanv: (Note that this is the backup server, so it will only be hit until SP itself is ready.)

PR #160, opened 2020-10-14
Allow jinja2 template inheritance

stefanv:
/cc @kmshin1397


kmshin1397 (approved):
Tested locally and seems to be work perfectly!

acrellin (approved):

PR #167, opened 2020-10-23
nginx: tweak buffer sizes for common upload/download scenario

stefanv:
In our test case (SkyPortal), users often upload thumbnails on the
order of 20k to 40k; we're setting the upload buffer to 64k to
accommodate those.

For typical JSON bodies, these tend to land under 256k, so we provide
adequate buffer space for those too.

Responses should be a bit faster if they stay within buffer
limits (nothing gets buffered to disk then).


dmitryduev (approved):
👍

acrellin (approved):

PR #173, opened 2020-10-28
Redirect webdriven firefox console logs to geckodriver logs

dannygoldstein:
This PR alters the `MyCustomWebDriver` class which is used in unit testing. It redirects the console logs from web browser instances driven by the driver to stdout, which is then rerouted to `geckodriver.log`. This allows access to the webdriver console from unit tests


stefanv (approved):
LGTM; merge conflict.

PR #175, opened 2020-10-29
Robustify test_utils.py clicking

kmshin1397:
This probably won't magically solve all the flakiness, but I think I found something that'll bring noticeable improvement.
- Move to an element after scrolling to it before attempting to click.
- Expose an option to have `scroll_to_element` scroll within the parent container instead of the window (this helps a lot with forms where we click an option from a long dropdown list). 
- Allow `StaleElementReferenceElement` to pass through the scripted clicking onto the ActionChains click attempt. I've noticed this error pop up occasionally on MUI components where the element comes up stale for some reason even though visually it is still there (specifically saw this today while looking at the classification delete button in `test_classifications`). 

I've tried out these changes locally and was able to get a number of problematic tests passing without any reruns (including `test_classifications` which we've XFAIL'ed for awhile now); will make a follow-up PR to SP once this is in to update those tests and any others I find that'll benefit from this.

### DIFF
--- a/app/test_util.py
+++ b/app/test_util.py
@@ -57,8 +57,7 @@ class MyCustomWebDriver(RequestMixin, webdriver.Firefox):
 
     def wait_for_css(self, css, timeout=10):
         return WebDriverWait(self, timeout).until(
-            expected_conditions.presence_of_element_located(
-                (By.CSS_SELECTOR, css))
+            expected_conditions.presence_of_element_located((By.CSS_SELECTOR, css))
         )
 
     def wait_for_xpath_to_appear(self, xpath, timeout=10):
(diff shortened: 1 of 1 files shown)


acrellin (approved):
Looks good. Considering the amount of style changes, I'd suggest we run `black` on this file for consistency w/ SP.

dannygoldstein (approved):

app/test_util.py
```diff
@@ -42,8 +43,10 @@ def server_url(self, value):
     def get(self, uri):
         webdriver.Firefox.get(self, self.server_url + uri)
         try:
-            self.find_element_by_id('websocketStatus')
-            self.wait_for_xpath("//*[@id='websocketStatus' and contains(@title,'connected')]")
+            self.find_element_by_id("websocketStatus")
```
stefanv, line 46: What was the goal of this line?

app/test_util.py
```diff
@@ -88,68 +87,75 @@ def wait_for_xpath_to_be_unclickable(self, xpath, timeout=10):
...
         except ElementClickInterceptedException:
             pass
+        except StaleElementReferenceException:
+            pass
 
         try:
             return self.execute_script("arguments[0].click();", element)
         except JavascriptException:
             pass
+        except StaleElementReferenceException:
```
stefanv, line 127: can probably use `except (JavascriptException, StaleElementReferenceException):`

stefanv (commented):
Nice, I'm glad this worked!

Curious whether scroll_parents is best left True or False by default.

PR #179, opened 2020-12-03
Use correct 503 code when service is unavailable

stefanv:


dmitryduev:
Lol. At least it wasn't 418!

PR #181, opened 2020-12-08
Refactor permissions and add an integer primary key for API access to join table rows 

dannygoldstein:


dannygoldstein:
Closing in favor of #184

PR #182, opened 2020-12-11
Hash JavaScript directory

acrellin:
See https://github.com/skyportal/skyportal/pull/1492


stefanv:
I don't think this is the right approach—we should be hashing in webpack.

PR #188, opened 2021-01-21
allow users to control which tests they run using make directives

dannygoldstein:
Currently when users call 

```
make test
```

or 

```
make test_headless
```

there is no way of specifying which tests to run; all tests are run. Recently it was requested that the new skyportal model tests be broken out to run in a separate CI job than the api and frontend tests to keep the runtime of CI down. In order to do this we need a way of specifying which tests to run in a `make` directive.  This PR facilitates that by allowing users to specify a `TEST_SPEC` to the `make test` and `make test_headless` directives. E.g., to run only the api tests via `make test`, one could do

```
make test TEST_SPEC='skyportal/tests/api'
```

and only the API tests would run. If `TEST_SPEC` is not specified then the current behavior is recovered, i.e., all tests run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,20 @@ SUPERVISORD_CFG=baselayer/conf/supervisor/supervisor.conf
 SUPERVISORD=$(PYTHON) -m supervisor.supervisord -s -c $(SUPERVISORD_CFG)
 SUPERVISORCTL=$(PYTHON) -m supervisor.supervisorctl -c $(SUPERVISORD_CFG)
 
+# Set the test spec to the empty string by default,
+# which causes the make test_headless and make test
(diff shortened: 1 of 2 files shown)


stefanv:
The way to run specific tests is via the test_frontend tool currently, as specified in the (SkyPortal) developer docs.

Make is annoying this way, in that it is hard to extract parameters (or you have to send them in with environment variables like you did here).

Is there a good reason why the tool cannot be invoked directly?  Python is much more flexible than bash.

dannygoldstein:
@stefanv you are absolutely right - that's a way better way to go. Doing it that way is possible without this PR and works fine (actually better as there are fewer escaping issues). Closing this PR in favor of that.

PR #191, opened 2021-01-27
Ensure config.yaml app.routes matches that of config.yaml.defaults (in parent app)

acrellin:
This PR addresses the issue where a local config.yaml falls behind config.yaml.defaults, leading to missing front-end route mappings, causing critical failures.

Closes https://github.com/skyportal/skyportal/issues/1531

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -116,7 +116,7 @@ def load_config(config_files=[]):
             "Warning: You are running on the default configuration. To configure your system, "
             "please copy `config.yaml.defaults` to `config.yaml` and modify it as you see fit."
         )
-    else:
+    elif os.path.exists(basedir / "../config.yaml"):
         ensure_yaml_structure_matches_defaults(
             basedir / "../config.yaml",
             basedir / "../config.yaml.defaults",
--- a/app/config.py
+++ b/app/config.py
@@ -10,7 +10,7 @@ from ..log import make_log
 log = make_log('baselayer')
 
 
-def ensure_yaml_structure_matches_defaults(config_path, defaults_config_path):
+def ensure_yaml_routes_matches_defaults(config_path, defaults_config_path):
     # Borrowed from https://github.com/dmitryduev/kowalski/blob/master/kowalski.py
     # check contents of config WRT config.yaml.defaults
     with open(defaults_config_path) as defaults_yaml:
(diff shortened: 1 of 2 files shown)


dmitryduev:
@acrellin weren't we using somewhere (on the CI?) a very minimalistic config? I could have dreamt it though...

dmitryduev (approved):

acrellin:
@dmitryduev test_config.yaml perhaps? That's very minimal, but this only checks the parent app's config.yaml (if specified) against its config.yaml.defaults

dmitryduev:
ah, right, never mind then.

app/config.py
```diff
@@ -1,13 +1,35 @@
...
 
 
+def ensure_yaml_structure_matches_defaults(config_path, defaults_config_path):
+    # Borrowed from https://github.com/dmitryduev/kowalski/blob/master/kowalski.py
+    # check contents of config WRT config.yaml.defaults
+    with open(defaults_config_path) as defaults_yaml:
+        config_defaults = yaml.load(defaults_yaml, Loader=yaml.FullLoader)
+    with open(config_path) as config_yaml:
+        config_wildcard = yaml.load(config_yaml, Loader=yaml.FullLoader)
+    deep_diff = DeepDiff(config_wildcard, config_defaults, ignore_order=True)
```
stefanv, line 20: I think we (also) want the option to check whether a config file complies with the schema.  For that case `ignore_order` must be set to False.

stefanv: @dmitryduev I presume this was written to ensure that all fields are present in the config, which is not a strict requirement for us?

acrellin: @stefanv it is WRT app.routes -- so perhaps I could change this to simply log a message if any other parts of the config don't match, but error if app.routes in config.yaml doesn't match the defaults (which will lead to missing route/component mappings and critical front-end crashes)

dmitryduev: @stefanv yes, that's a requirement on Kowalski, but not here, so I support @acrellin's suggestion.

acrellin: Done

stefanv (approved):

dmitryduev:
@acrellin failure related - `deepdiff` missing in requirements:
> ModuleNotFoundError: No module named 'deepdiff'

dmitryduev (changes requested):
@acrellin failure related - `deepdiff` missing in requirements:
> ModuleNotFoundError: No module named 'deepdiff'

acrellin:
Woops, thanks for catching that. Added `deepdiff` to requirements.txt now

dmitryduev:
@acrellin the CI saw a walrus and freaked out -- gotta change the python version requirement on Travis. Or better still, ditch it for good as @dannygoldstein is suggesting and move to GA asap.

stefanv:
> @acrellin the CI saw a walrus and freaked out -- gotta change the python version requirement on Travis. Or better still, ditch it for good as @dannygoldstein is suggesting and move to GA asap.

I'm on it.

stefanv:
@dmitryduev I think this is probably good to go.

dmitryduev (approved):
Forgot about this one, my bad. LGTM!

dmitryduev:
@stefanv @acrellin Hmm.. the CI failed on commit to master?..

PR #193, opened 2021-01-28
Add custom jinja filters

stefanv:
Needed to add the version to the JS bundle

### DIFF
--- a/tools/fill_conf_values.py
+++ b/tools/fill_conf_values.py
@@ -50,8 +50,7 @@ def fill_config_file_values(template_paths):
             env = jinja2.Environment(
                 loader=jinja2.FileSystemLoader(tpath),
             )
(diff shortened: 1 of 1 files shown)


tools/fill_conf_values.py
```diff
@@ -30,7 +50,8 @@ def fill_config_file_values(template_paths):
             env = jinja2.Environment(
                 loader=jinja2.FileSystemLoader(tpath),
             )
-            env.filters['md5sum'] = md5sum
+            for (name, func) in custom_filters.items():
+                env.filters[name] = func
```
acrellin, line 54:
```suggestion
            env.filters.update(custom_filters)
```
Any reason not to use this simpler form?

stefanv: No, that's good!

acrellin (approved):

PR #195, opened 2021-02-19
Add new CronJobRun model; add instances when cron jobs are executed

acrellin:


acrellin:
@stefanv this is the PR I was referring to -- whenever you have a moment, I'd love a review

app/models.py
```diff
@@ -1157,3 +1157,17 @@ def is_readable_by(self, user_or_token):
...
+
+
+class CronJobRun(Base):
+    """A record of a run (or attempted run) of a cron job."""
+    script = sa.Column(
+        sa.String,
+        nullable=False,
+        doc="Name of script being run.",
+    )
+    exit_status = sa.Column(
```
stefanv, line 1169: Should we also log stdout/stderr?

acrellin: @stefanv  Good call - just pushed up that change

services/cron/cron.py
```diff
@@ -78,10 +80,18 @@ def cache_to_file(self):
...
             tc.reset(key)
             try:
-                subprocess.Popen(
+                proc = subprocess.Popen(
                     script, stdout=subprocess.PIPE, stderr=subprocess.STDOUT
                 )
+                output, _ = proc.communicate()
             except Exception as e:
                 log(f'Error executing {script}: {e}')
+                DBSession().add(CronJobRun(script=script, exit_status=1))
```
stefanv, line 89: Add the exception string here as the proc_output?

stefanv (approved):
I suggest adding the exception string, if we don't get output.

Go ahead and merge if you're happy with that change and have made it.

app/models.py
acrellin: I tested that it captures both stdout and stderr

services/cron/cron.py
acrellin: Done - renamed proc_output to output, and just store the exception string there in that case

acrellin:
Thanks, @stefanv !

PR #197, opened 2021-02-25
Monitor template files

stefanv:
I sometimes forget that `.template` files aren't compiled automatically, prompting investigation into why changes aren't propagated.

This PR introduces a file monitor, that compiles templates when they are modified, which in turn triggers webpack (in the case of JavaScript).

### DIFF
--- a/services/template_watch/template_watch.py
+++ b/services/template_watch/template_watch.py
@@ -1,0 +1,46 @@
+import os
+from os.path import join as pjoin
+import time
+import sys
+
+from watchdog.observers import Observer
+from watchdog.events import PatternMatchingEventHandler
+
+from baselayer.log import make_log
+from baselayer.tools.fill_conf_values import fill_config_file_values
+from baselayer.app.env import load_env
+
+log = make_log('template_watch')
+env, cfg = load_env()
+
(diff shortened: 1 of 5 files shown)


services/template_watch/template_watch.py
```diff
@@ -0,0 +1,46 @@
...
+# Baselayer parent project path
+watchpath = os.path.abspath(pjoin(os.path.dirname(__file__), '../../..'))
+
+
+class EventHandler(PatternMatchingEventHandler):
+    def __init__(self):
+        PatternMatchingEventHandler.__init__(self, patterns=['*.template'])
+
+    def on_modified(self, event):
+        fill_config_file_values([event.src_path])
```
dmitryduev, line 29: Maybe log which template changed and triggered this?

dmitryduev (approved):

services/template_watch/template_watch.py
stefanv: It's already logged by the template compiler.

acrellin (approved):
Nice, thanks! Yes, this'll be handy to not have to bring the app down and back up again when editing .template files

PR #201, opened 2021-03-04
Do not log 404s from HEAD requests

acrellin:
Closes https://github.com/skyportal/skyportal/issues/1164

### DIFF
--- a/app/handlers/base.py
+++ b/app/handlers/base.py
@@ -259,7 +259,8 @@ class BaseHandler(PSABaseHandler):
         extra : dict
             Extra fields to be included in the response.
         """
-        log(f'Error response returned by [{self.request.path}]: [{message}]')
+        if not (status == 404 and self.request.method == "HEAD"):
+            log(f'Error response returned by [{self.request.path}]: [{message}]')
 
         self.set_header("Content-Type", "application/json")
         self.set_status(status)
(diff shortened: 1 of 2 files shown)


stefanv:
Is there a matching BL part to this?

acrellin:
@stefanv this *is* a BL PR 😉

stefanv:
:joy: OK, then I think it's not quite done—the majority of log calls go into the nginx logs?

acrellin:
@stefanv :)  I'm not sure what you mean

stefanv:
I think the log entries are 404 HEAD entries made by nginx.

acrellin:
@stefanv you mean messages like this? `WARNING:tornado.access:400 PUT /api/comment/26238 (127.0.0.1) 14.89ms`

stefanv:
Yes, exactly.

acrellin:
nginx logging level is currently at "info" -- should we bump that to "error" or at least "warn"?

stefanv:
Probably; but I think we'll still run into the 404 HEADs that way.

acrellin:
Nice, there's already a conditional logging check, so I've just updated that to not log 404s -- wanna take another look @stefanv ?

stefanv:
It must log 404's, just not HEAD calls.

stefanv:
:rocket:

PR #204, opened 2021-03-15
add logical not to UACL ops

dannygoldstein:
This PR adds the functionality to create new `UserAccessControl`s that are the logical NOT of other `UserAccessControl`s. 

This is useful for defining access controls that only allow access if a certain condition is NOT met. 

The implementation here uses an anti join against the predicate to be negated.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -132,6 +132,25 @@ class UserAccessControl:
 
         raise NotImplementedError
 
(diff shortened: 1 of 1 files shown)


dannygoldstein:
This is actually not needed for what I am trying to do, so closing.

PR #205, opened 2021-03-15
Temporarily require sqlalchemy <1.4.0 to make compatible with sqlalchemy-utils

acrellin:
Open `sqlalchemy-utils` issue notes incompatibility w/ `sqlalchemy` versions >=1.4.0. Until they release a fix, we'll avoid the breaking versions explicitly here.

See https://github.com/kvesteri/sqlalchemy-utils/issues/505


RyanPaulMcKenna:
I amtrying to use "from sqlalchemy_utils import Timestamp". I am using Flask-SQLAlchemy version 2.4.1. please advise which if any version of SQLAlchemy-utils I will be able to use with Flask-SQLAlchemy. Alternatively, does anyone know an alternative means of Timestamping a user class object?

stefanv:
@RyanPaulMcKenna I'm a bit confused: are you using Flask-SQLAlchemy in combination with baselayer?  Baselayer is pretty tightly coupled to Tornado.

According to the Flask-SQLAlchemy source, they require SQLAlchemy >= 1.2, so any 1.2 or 1.3 version should do the trick.

PR #208, opened 2021-03-19
Shut down all webpack processes

stefanv:
Otherwise, webpack leaves a node server running for about 40 seconds


services/webpack/supervisor.conf
```diff
@@ -7,3 +7,4 @@ startretries=0
 startsecs=1
 stdout_logfile=log/webpack.log
 redirect_stderr=true
+stopasgroup=True
```
acrellin, line 10: `true`?

stefanv: Updated.

acrellin (approved):

PR #210, opened 2021-03-19
Add service for mocking external APIs in testing

kmshin1397:
This PR adds a `test_service` service to Baselayer which can be used to mock external API requests during testing. The idea is to tweak configurations for test instances of a Baselayer application (through a `test_config.yaml` like in SkyPortal, for example) such that external API calls made by the application's handlers are directed to `http://localhost:{configs[ports.test_server]}` instead of the real API host; the `test_service` will return mock responses to the Baselayer application. These mock responses are set up and customized by passing to the microservice via configurations the path to a Python module which exposes a `make_app()` function returning a  `tornado.web.Application` instance. An example such module is provided and configured in the default configs.

This is a precursor to addressing skyportal/skyportal#1792. A follow-up PR will be made to SkyPortal with a more sophisticated test server module to instantiate the microservice with in order to mock external calls within the SkyPortal test suite.

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -67,6 +67,7 @@ ports:
     websocket_path_out: 'ipc://run/message_flow_out'
     status: 64500
     migration_manager: 64501
+    test_server: 64502
 
 external_logging:
     papertrail:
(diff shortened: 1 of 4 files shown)


acrellin (approved):

stefanv (commented):
@kmshin1397 This is the right idea, but I wonder if it's slightly overkill.  Baselayer allows you to add a `services` directory in your project, as described here: http://cesium-ml.org/baselayer/usage.html#microservices

In other words, it is not necessary to get a factory from the config and file that up: the service can simply be added under `skyportal_repo/services` and run itself on whatever port it wants and start serving.  We don't even need support from baselayer to do that.

kmshin1397:
@stefanv Thanks for pointing that out. I'd missed that functionality for detecting services in a `skyportal/services` directory. I agree that this PR seems overkill now - closing in favor of just a SkyPortal PR.

PR #222, opened 2021-04-14
Make database engine creation defaults adjustable.

stefanv:
This needs to be followed up with a change to SkyPortal, where the connections are bumped up to 10 + 10 (instead of 5 + 10, as it is now).


acrellin (approved):
Looks good!

PR #225, opened 2021-05-03
Simplify access controls for system admins

kmshin1397:
This PR simplifies the permissions querying for system admins for the baselayer-provided permission rule classes. Specifically, all rows are just returned if the user is a system admin without doing any of the normal checks when using the `AccessibleIfRelatedRowsAreAccessible` and `ComposedAccessControl` classes. This reflects the same approach already used in `AccessibleIfUserMatches`. 

For example, if you have a rule that is the AND of two rules, the `ComposedAccessControl` as is will join the entire table in question to itself since both pieces of the permission rule will return all rows for a system admin. Now, we'll just return the table once.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -68,8 +68,13 @@ utcnow = func.timezone("UTC", func.current_timestamp())
 # The db has to be initialized later; this is done by the app itself
 # See `app_server.py`
 def init_db(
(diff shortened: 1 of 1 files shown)


acrellin (approved):
I had thought we were already doing this. Good catch!

PR #226, opened 2021-05-04
Return 404 when requested resource not found/inaccessible

acrellin:


app/handlers/base.py
```diff
@@ -309,6 +309,8 @@ def error(self, message, data={}, status=400, extra={}):
         if not (status == 404 and self.request.method == "HEAD"):
             log(f'Error response returned by [{self.request.path}]: [{message}]')
 
+        if message.startswith("Access Error: Invalid"):
```
stefanv, line 312: Should this live in `write_error` instead?

acrellin: I don't think it matters -- `write_error` just calls `error`. If we were just checking for the `AccessError` exception type it'd make more sense to have it there, but we don't want to return 404 for all `AccessError`s raised -- just those with this error message (there are other cases where it's raised where 404 doesn't make as much sense).

stefanv: That's a very specific thing to catch and handle in a special way.  Is this generic enough?  Is there a way to make it more generic (such as raising an ObjectNotFound exception and handling that differently?).

acrellin:
Sure, that sounds good. Will update.

acrellin:
@stefanv updated, ready for re-review

stefanv:
I like it!

PR #227, opened 2021-05-12
Update record accessibility check

acrellin:
This commit https://github.com/skyportal/skyportal/pull/1982/commits/091263665c28c78a19e305676f7b2b0ec590bf79 causes `is_accessible_by` to error for `StreamUser` records without this change

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -845,6 +845,8 @@ class BaseMixin:
         # Query for the value of the access_func for this particular record and
         # return the result.
         result = accessibility_table.scalar()
+        if result is None:
+            result = False
 
         if not isinstance(result, bool):
             raise RuntimeError(


dannygoldstein:
I'm not sure this should ever be necessary ?

dannygoldstein:
Proposing an alternative ...

dannygoldstein:
Actually, before I do, I am having trouble understanding this error case. Could you help me wrap my head around it?

If I do, in postgres, 

`skyportal=# select count(*) > 0  from photometry where false;`

the result is 

```
 ?column?
----------
 f
(1 row)
```

can you show the SQL query that results in this being none (0 rows in SQL output)?

acrellin:
@dannygoldstein did you see the linked commit? https://github.com/skyportal/skyportal/commit/091263665c28c78a19e305676f7b2b0ec590bf79 

With that commit, `is_accessible_by` errors for StreamUser records

dannygoldstein:
I couldn’t pull that commit, seems like it is not part of the repo ?

________________________________

acrellin:
Yeah, it's not merged to SP yet -- you'll have to fetch my fork first

acrellin:
@dannygoldstein see https://github.com/skyportal/skyportal/pull/1982

kmshin1397:
@dannygoldstein Here is the SQL generated by SQLAlchemy which generated the 0 row output:

```
SELECT          count(*) > 0 AS delete_ok
FROM            stream_users
JOIN            users
ON              users.id = stream_users.user_id
LEFT OUTER JOIN (group_users AS group_users_1
              JOIN            groups
              ON              groups.id = group_users_1.group_id)
ON              users.id = group_users_1.user_id
LEFT OUTER JOIN group_streams
ON              group_streams.group_id = groups.id
AND             group_streams.stream_id = stream_users.stream_id
WHERE           false
AND             stream_users.id = 1
GROUP BY        stream_users.id
HAVING          bool_and(group_streams.stream_id IS NULL)
```

kmshin1397:
Looks like the GROUP BY/HAVING is what causes it to return 0 rows instead of the false (corresponding to [these lines](https://github.com/skyportal/skyportal/blob/54a5dab8c89815b988e260f6613e72150782ed95/skyportal/models.py#L3526-L3529))

kmshin1397 (approved):

acrellin:
Thanks for digging into this a bit more deeply, @kmshin1397

acrellin:
@dannygoldstein wanna take another look / offer any feedback before we merge? See Kyung's explanatory comments above.

dannygoldstein:
I would consider moving the inner query to a subquery and selecting count(*) > 0 from that

________________________________

acrellin:
Closing as no longer needed

PR #229, opened 2021-05-18
Simplify access controls for system admins

acrellin:


acrellin:
Let's not merge this into master until we get the necessary patches in place in SP

app/models.py
```diff
@@ -474,6 +481,12 @@ def query_accessible_rows(self, cls, user_or_token, columns=None):
             Query for the accessible rows.
         """
 
+        # system admins automatically get full access
+        if user_or_token.is_admin:
+            return public.query_accessible_rows(
+                cls, user_or_token, columns=columns
+            )
+
```
dannygoldstein, line 489: we should not do this, because it will allow admins to violate restrictions that should apply even to them. there are a few that are used now to enforce data model integrity. if we merge this change, we will no longer have a guarantee of these data model assumptions: 

A user can only join a group if they have access to all of the group's streams.
A stream can only be removed from a group if none of the Group's filters currently filter the stream.
A stream can only be added to a group if all of the group's members have access to the stream.
A group user's admin status can only be modified by group admins.
Only a group admin or the target user can remove a user from a group.
Only system admins can grant and revoke stream access to users.
Only a group admin can add a user to a group.
Users cannot add streams to single user groups.
No one can remove a user from their single user group (this can only be done by an event listener).
No one can add a user to a single user group through the groups api.

This may lead to subtle but critical errors down the line that become hard to debug

dannygoldstein: Example: dangling filter reference

I have a group with access to streams 1, 2 and a filter filtering both streams. I, as an admin that can access anything, decide to remove stream 2 from the group. However, the filter is still filtering stream 2. There is now a dangling reference, where everyone in group 2 is now accessing candidates based off of a filter that is filtering photometry that they do not have access to

acrellin:
cc @kmshin1397 ^ (the author of this PR)

app/models.py
kmshin1397: Ok you're right @dannygoldstein I'd forgotten to consider that. What are your thoughts on updating this PR so that the `ComposedAccessControl` returns the full table for system admins only if a `CustomUserAccessControl` is not part of the chain of policies to join together? I believe all those special cases you mentioned where we want to limit system admin access are implemented via `CustomUserAccessControl` rules. 

I would still like to solve the original issue I saw that this PR was meant to address, which is that if you have a rule that is the AND of two rules, the ComposedAccessControl as is will join the entire table in question to itself since both pieces of the permission rule will return all rows for a system admin, leading to slower than necessary queries.

dmitryduev:
Stalled, closing as we don't have too many sysadmins that could be affected by this :)

PR #230, opened 2021-05-18
Return 404 when requested resource not found/inaccessible

acrellin:


acrellin:
Let's not merge this until we get the necessary updates in place in SP

acrellin:
Woops, botched a rebase and overwrote the original diff here.. Will update.

acrellin:
@stefanv @kmshin1397 Updated, ready for review

app/handlers/base.py
```diff
@@ -309,6 +310,8 @@ def error(self, message, data={}, status=400, extra={}):
         if not (status == 404 and self.request.method == "HEAD"):
             log(f'Error response returned by [{self.request.path}]: [{message}]')
 
+        if message.startswith("Access Error: Invalid"):
```
stefanv, line 313: Is this left over from the previous version?

acrellin: Ah yes, thanks,willupdate

acrellin: Updated

kmshin1397 (approved):

acrellin:
@stefanv how does this look to you now?

PR #233, opened 2021-06-23
Add expiration_date to Users

kmshin1397:
Also changed the `is_active` function for Users, which currently is just returning `True` no matter what, so that it serves as a check for expired users. We don't seem to be using this function anywhere, but doesn't hurt to have it be more of a real function in case we use it later. This will be accompanied by a PR on SkyPortal to actually use the new column for user expiration functionality (setting expired users to view only automatically). Part of addressing fritz-marshal/fritz#180

Plus black formatting.

### DIFF
--- a/app/access.py
+++ b/app/access.py
@@ -26,13 +26,13 @@ def auth_or_token(method):
             if token is not None:
                 self.current_user = token
                 if not token.created_by.is_active():
-                    raise tornado.web.HTTPError(403, "User account expired")
+                    raise tornado.web.HTTPError(401, "User account expired")
             else:
-                raise tornado.web.HTTPError(403)
+                raise tornado.web.HTTPError(401)
             return method(self, *args, **kwargs)
         else:
             if not self.current_user.is_active():
-                raise tornado.web.HTTPError(403, "User account expired")
+                raise tornado.web.HTTPError(401, "User account expired")
             return tornado.web.authenticated(method)(self, *args, **kwargs)
 
     return wrapper
@@ -51,7 +51,7 @@ def permissions(acl_list):
                 set(acl_list).issubset(self.current_user.permissions)
                 or "System admin" in self.current_user.permissions
             ):
-                raise tornado.web.HTTPError(403)
+                raise tornado.web.HTTPError(401)
             return method(self, *args, **kwargs)
 
         return wrapper
--- a/app/access.py
+++ b/app/access.py
@@ -26,13 +26,13 @@ def auth_or_token(method):
             if token is not None:
                 self.current_user = token
                 if not token.created_by.is_active():
-                    raise tornado.web.HTTPError(401, "User account expired")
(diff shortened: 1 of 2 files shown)


acrellin (approved):

acrellin:
This all looks good, but one note for the SP PR: it feels more intuitive to have a user expiration date and the active/inactive flag correspond to deactivating the account altogether rather than just making them view only. But I guess that'll have to be decided by the PIs.

kmshin1397:
> This all looks good, but one note for the SP PR: it feels more intuitive to have a user expiration date and the active/inactive flag correspond to deactivating the account altogether rather than just making them view only. But I guess that'll have to be decided by the PIs.

Ah yeah, that makes sense and we can discuss that further. I think the view only way was the easier path of implementation, since I imagine completely deactivating accounts will mean implementing a new "Deactivated accounts" table for the user management page, maybe blocking inactive accounts from logging in, canceling their API tokens, etc. I just saw less steps to simply setting them to view only but definitely wiling to change directions here.

kmshin1397:
Instead of adding another flag column, I just used the existing `User.is_active()` function mentioned above to use to guard against expired users doing anything on the API. This guard logic has been incorporated into `access.py`

acrellin (approved):
Nice!

acrellin (changes requested):
Ah just realized we'll also need to add the check to the `permissions` decorator here, especially since we'll no longer be removing ACLs

acrellin (approved):
Haha, my bad, I see that permissions is already decorated with auth_or_token!

app/access.py
```diff
@@ -19,16 +19,20 @@ def auth_or_token(method):
...
+        token_header = self.request.headers.get("Authorization", None)
+        if token_header and token_header.startswith("token "):
+            token_id = token_header.replace("token", "").strip()
             token = Token.query.get(token_id)
             if token is not None:
                 self.current_user = token
+                if not token.created_by.is_active():
+                    raise tornado.web.HTTPError(403, "User account expired")
             else:
                 raise tornado.web.HTTPError(403)
```
stefanv, line 31: Should this be 401?

stefanv (approved):
Looks good.

I see we used `403` before where we probably should have used `401`.  Do you mind fixing that in this PR?

app/access.py
kmshin1397: Agreed, this has been changed

stefanv:
@kmshin1397 Sorry, I may not have been clear: *Your* implementation was correct, we just needed to switch the code on the *old* one.  I.e., 401 if the token is not recognized and 403 if the token is valid (user is recognized) but access is forbidden.

kmshin1397:
Ah right okay yes I see the difference between the two now. _Now_ it should be fixed

stefanv:
Thanks, Kyung!

PR #236, opened 2021-07-14
Nginx-based rate limiting

acrellin:


services/nginx/nginx.conf.template
```diff
@@ -5,6 +5,19 @@ pid run/nginx.pid;
 http {
   sendfile on;
 
+  geo $limit {
```
stefanv, line 8: This seems to mean "requests from localhost do not get rate limited".  Is that the intent?

stefanv (approved):

services/nginx/nginx.conf.template
acrellin: That's correct

acrellin: Without this, `make load_demo_data` fails

acrellin: Eventually we can template this to add Kowalski's IP address (in fritz-deploy)

kmshin1397 (approved):

PR #238, opened 2021-07-16
Further parameterize rate limit settings in config.yaml.defaults

acrellin:


acrellin:
@stefanv self-merging this small patch to keep things moving today, but happy to amend/revert if you have any issues with it

stefanv:
👍

PR #239, opened 2021-07-20
Include truncated http_authorization in nginx log messages

acrellin:
Log messages will now be of the form `127.0.0.1 - - [token e7a82026] [20/Jul/2021:14:59:50 -0700] "GET /api/sysin HTTP/1.1" 400 110 "-" "python-requests/2.25.1"`

### DIFF
--- a/services/nginx/nginx.conf.template
+++ b/services/nginx/nginx.conf.template
@@ -57,7 +57,7 @@ http {
       default 1;
   }
 
-  log_format elb_log '$proxy_add_x_forwarded_for - $remote_user [$time_local] ' '"$request" $status $body_bytes_sent "$http_referer" ' '"$http_user_agent"';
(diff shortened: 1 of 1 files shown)


stefanv:
@acrellin What do the log entries look like now?  I presume they contain `Authorization: token foo`.

Can we do a regex replacement to get something similar to `token=foo` in the log instead?

Here's an example: https://serverfault.com/a/1006190

acrellin:
@stefanv the format is now `127.0.0.1 - - token e7a82026-1446-49ee-8bb7-44f2bd94b8e3 [20/Jul/2021:13:02:49 -0700] "GET /api/sysin HTTP/1.1" 400 110 "-" "python-requests/2.25.1"` which I quite like. I propose we leave it as is.

stefanv:
OK, just place it in square brackets and I'll merge.

PR #240, opened 2021-07-22
Update record accessibility check

acrellin:
Re-opening as this is needed to address a bug when calling `get_if_accessible_by` w/ `raise_if_none=True` for some models.


acrellin:
The prior PR associated with this patch was already approved, so merging.

PR #243, opened 2021-07-28
Use nginx real_ip module to uncover originating IP

stefanv:


stefanv:
Untested, but this captures everything.

acrellin (approved):
It was that simple, huh? Nice! Did you test manually on Staging?

stefanv:
Yes, I left the config in preview in case you wanted to take a look. I think they should render similarly, once we fill out the IPs in the config.

PR #245, opened 2021-07-29
Dedicate some processes to frontend only

stefanv:


profjsb (commented):
Hard coding fail timeouts of 5s in both. Seems like we could be more lenient now on the frontend side, yes?

stefanv:
@profjsb The reason the timeout is 5s is because nginx annoyingly makes the timeout and the out-of-service time the same.  So, if after 5s a service is not available, it now waits 5s.  We had that on 30s, but that means that if the process is not available at any point, nginx waits 30s before re-entering it into the pool.

services/nginx/nginx.conf.template
```diff
@@ -35,20 +35,36 @@ http {
     server localhost:{{ ports.fake_oauth }};
   }
 
-  upstream frontends {
+  upstream frontend {
     least_conn;
     {% for p in range(server.processes) -%}
```
acrellin, line 40: Doesn't this now need to become `server.dedicated_frontend_processes`?

stefanv: No, so the way it works is that the frontend gets access to all processes, but the API is restricted to a subset only.

stefanv: In principle we could fully isolate them, but I don't think it's necessary.  If a frontend request happens to go to one of the API processes, that's fine.  Just want to make sure some are reserved specifically for frontend too.

stefanv: Basically, API always is second priority to serve up.  Which sounds weird, but I think is the right call for usability.

acrellin (approved):

PR #251, opened 2021-09-02
Satisfy joint dependencies between all requirement.txt files

stefanv:
We are currently using `pkg_resources.require` to ensure that
dependencies are met.  However, and perhaps recently, this function
only checks that the dependencies can be resolve, *not that it is
resolved by the currently installed distribution*. We switch to
`pkg_resources.working_set.resolve` which does exactly that.

If requirements are not met, we call pip to install the missing
packages. Before, that was done on each `requirements.txt` file
separately, which does not ensure that the joint set of dependencies
is met.  This is fixed by calling `pip` with multiple `-r
requirements_file` arguments.

Before, when there were version conflicts in the requirements
specification, we called pip to try and resolve it.  My understanding
is that this is a fatal error (e.g., requiring `requests<25` and
`requests>25` simultaneously).  I therefore print out the error and abort.

### DIFF
--- a/tools/pip_install_requirements.py
+++ b/tools/pip_install_requirements.py
@@ -1,5 +1,5 @@
 import pkg_resources
-from pkg_resources import DistributionNotFound, VersionConflict
+from pkg_resources import DistributionNotFound, VersionConflict, Requirement
 
 import sys
 import subprocess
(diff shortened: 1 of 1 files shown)


profjsb:
Looks good to me. Works with my current branch of SP.

PR #253, opened 2021-09-14
Return HTTP 401 (Unauthorized) if token authentication fails

stefanv:
Currently we return HTTP 400 (Bad Request), which does not match an authorization failure.

### DIFF
--- a/app/handlers/base.py
+++ b/app/handlers/base.py
@@ -20,6 +20,7 @@ import social_tornado.handlers as psa_handlers
 from .. import psa  # noqa
 
 from ..models import DBSession, User, handle_inaccessible
+from ..custom_exceptions import AccessError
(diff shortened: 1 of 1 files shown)


stefanv:
@acrellin I'd appreciate it if you could help me think this through a bit more.  I've been looking at why 400's were being propagated before.  Token checking *should* raise a 401, both in the case where the token is invalid or expired (this is in `access.py`).  AccessError, after successful authentication, should then use 403 (Forbidden).

acrellin:
@stefanv Could it be because there was no token header provided? If so, I just patched that here: https://github.com/cesium-ml/baselayer/pull/254

stefanv:
This happened when I used an invalid token (I modified the data_loader to use `token + "f"`.)

PR #254, opened 2021-09-14
Handle case where no Authorization header is provided

acrellin:


app/access.py
```diff
@@ -34,6 +34,8 @@ def wrapper(self, *args, **kwargs):
             if self.current_user is not None:
                 if not self.current_user.is_active():
                     raise tornado.web.HTTPError(403, "User account expired")
+            else:
+                raise tornado.web.HTTPError(403, "Credentials missing or malformed")
```
stefanv, line 38: This should be a 401. Can we provide more explicit instructions to the user?  Something like: `Credentials malformed; expected header of form "authorization: token abc123"`

stefanv: We could probably catch this more explicitly on line 24 (`token_header.replace`)?

acrellin:
@stefanv updated (though not sure what you're suggesting with the latter comment)

stefanv:
I just meant that we are doing a generic catch, instead of checking the form of the header explicitly.

acrellin:
Ah, don't lines 22-23 do just that?

stefanv:
Hrm, yes, you're right; so what is this error that we are catching?  How will it typically arise?

stefanv:
Thanks!

PR #255, opened 2021-09-29
Add nginx log file

lpsinger:
This makes it easier to diagnose nginx configuration errors. See skyportal/skyportal#2292.


acrellin:
@lpsinger we already have nginx logs, e.g. https://github.com/cesium-ml/baselayer/blob/master/services/nginx/nginx.conf.template#L1

PR #256, opened 2021-09-29
Add system dependency instructions for MacPorts

lpsinger:
I found out the hard way that you have to install nginx with the `+realip` variant. See skyportal/skyportal#2292.


acrellin (approved):

acrellin:
Thanks @lpsinger !

PR #260, opened 2021-10-18
Add TypeScript service

acrellin:


stefanv:
With the two-step process, do we have a guarantee that the webpack service will run late enough to catch the transpiled sources?  If not, we should write out a sentinel file to let that service know that tsc is done.

acrellin:
@stefanv The bundle size will be unaffected by combining `--mode=production` and `--watch`, we'll just be simply keeping the daemon alive to watch for any new changes (e.g. when `tsc` completes)

stefanv:
Yeah, not a big deal I think.

stefanv:
Looks like the test suite on master is unhappy :thinking:

PR #261, opened 2021-10-21
Refactor models

stefanv:


stefanv:
The real test will be against SkyPortal, I suppose.

app/handlers/base.py
```diff
@@ -1,11 +1,16 @@
+__all__ = ['BaseHandler']
```
acrellin, line 1: This shouldn't be necessary as we never do `from baselayer.app.handlers.base import *`

app/models/__init__.py
```diff
@@ -0,0 +1,20 @@
+from .access import *
+from .acl import ACL, Role
+from .base import *
+from .cron_job_run import CronJobRun
+from .join import JoinModel, join_model
+from .session import *
+from .token import *
+from .user import User
+from .util import *
```
acrellin, line 9: Why not use a consistent pattern here -- either import each item explicitly, or always use `*`?

app/models/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+# Define these here to prevent circular imports elsewhere
+RoleACL = join_model("role_acls", Role, ACL)
+RoleACL.__doc__ = "Join table class mapping Roles to ACLs."
+
+UserACL = join_model("user_acls", User, ACL)
+UserACL.__doc__ = "Join table mapping Users to ACLs"
+UserRole = join_model("user_roles", User, Role)
+UserRole.__doc__ = "Join table mapping Users to Roles."
+
+del access, acl, base, cron_job_run, join, session, token, user, util
```
acrellin, line 20: Why is this line needed?

stefanv: Probably because of the `*` imports that I should remove above.

guynir42:
@stefanv is this done? I think on SP it was completed a while ago...

stefanv:
No, we ran into difficulties we could not debug at the time. Will have to re-attempt.

PR #262, opened 2021-10-22
Add (user_id, name) Token unique constraint

acrellin:
This patch removes the table-wide unique constraint on
Token.name, and instead adds a new (user_id, token_name)
unique constraint so that different users can have tokens
of the same name.

This will require a migration when upgrading SkyPortal.

Closes https://github.com/skyportal/skyportal/issues/2195


acrellin:
@stefanv This is a small one; mind taking a look when you have a moment?
Thanks!

PR #267, opened 2021-12-21
Aliasing like that does not work for sqlalchemy >= 1.4

mcoughlin:
Allows the permissions checking to pass in sqlalchemy >= 1.4.

See: https://docs.sqlalchemy.org/en/14/changelog/migration_14.html
in particular: Built-in FROM linting will warn for any potential cartesian products in a SELECT statement

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -621,7 +621,7 @@ class ComposedAccessControl(UserAccessControl):
                 continue
 
             # use an alias to avoid name collisions.
-            target_alias = sa.orm.aliased(cls)
+            # target_alias = sa.orm.aliased(cls)
 
             # join against the first access control using a subquery. from a
             # performance perspective this should be about as performant as
@@ -629,7 +629,7 @@ class ComposedAccessControl(UserAccessControl):
             # name collisions. The subquery is automatically de-subbed by
             # postgres and uses all available indices.
             accessible = access_control.query_accessible_rows(
-                target_alias, user_or_token, columns=[target_alias.id]
+                cls, user_or_token, columns=[cls.id]
             ).subquery()
 
             # join on the FK
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ simplejson>=3.17.2
 requests>=2.25.1
 selenium>=3.141.0
 pytest>=5.4.3
-sqlalchemy>=1.3.7,<1.4.0
+sqlalchemy>1.4.0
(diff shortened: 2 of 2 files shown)


mcoughlin:
@stefanv @profjsb I don't have workflow powers.

profjsb:
> @stefanv @profjsb I don't have workflow powers.

Done. And added you with enhanced powers to this repo.

app/models.py
```diff
@@ -1363,6 +1366,7 @@ class Token(Base):
     name = sa.Column(
         sa.String,
         nullable=False,
+        unique=True,
```
profjsb, line 1369: I think this needs to get removed -- Token names can be identical for different users.

app/models.py
```diff
@@ -1385,8 +1389,6 @@ def is_readable_by(self, user_or_token):
         """
         return user_or_token.id in [self.created_by_id, self.id]
 
-    __table_args__ = (sa.UniqueConstraint("created_by_id", "name", name="token_name_userid_unique_constraint"),)
```
profjsb, line 1388: reinstate: we need to create a unique constraint on id+name.
See https://github.com/cesium-ml/baselayer/pull/262/files

profjsb (changes requested):
Unfortunately the linting made this PR look longer than it really is.  My main question is whether using `sa.select` in place of `DBSession().query` could cause leakage across users since the query will not be done in scoped session. Or perhaps the role of this change is only to facilitate the `construction` of the query itself, not the issuance of the query (which must be done in a scoped session). @stefanv thoughts?

Please revert the token constraint change as noted in the inline comments.

mcoughlin:
@stefanv https://github.com/cesium-ml/baselayer/pull/267/files#diff-90c680eba43456b516da8b4c2573d467ae17d1b0ed4373549f2a593ced3616d5L623
These lines here.

mcoughlin:
@profjsb Minimal PR is now here

profjsb:
merging this into it's own branch on cesium-ml/baselayer so we can try it out on SP.

stefanv:
Do we know whether it is safe to remove this? There's a comment abusive it which states its purpose, so would be good to understand its original intent.

mcoughlin:
Reading the 1.3 to 1.4 docs it seems like they made some changes to how aliases were handled, but presumably Danny needs to comment why it was there in the first place.

PR #269, opened 2022-01-14
Use psql for all database operations

stefanv:
Previously, we used dropdb, createdb, and createuser, but these
commands operate on the local database, and thus is not suitable for
working in a Docker container network.

### DIFF
--- a/tools/db_init.py
+++ b/tools/db_init.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-
 import subprocess
 import sys
 import argparse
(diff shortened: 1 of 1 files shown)


stefanv:
I need to test this inside skyportal still.

stefanv:
Tested locally at least, and seems to work OK.

profjsb (approved):
LGTM. Will merge and we can try it out in GHA/SP.

Theodlz:
Done ! @stefanv you can find it here: https://github.com/skyportal/skyportal/pull/2938

PR #270, opened 2022-01-15
Improve logging when creating and connecting to DB

stefanv:


tools/db_init.py
```diff
@@ -132,3 +139,5 @@ def test_db(database):
            {test_cmd + db}
         '''))
     sys.exit(1)
+
+print()
```
profjsb, line 143: what does this print statement do?

stefanv: It was just for some spacing in the log so this is more clearly delineated. Not super important, feel free to remove.

profjsb:
Is this ready to merge @stefanv ?

stefanv:
Yes, please.

PR #279, opened 2022-01-25
No longer enforce a route structure on the config file

stefanv:
Before, we forced config.yaml to reflect the routes in
config.yaml.default.  However, users may want to add or remove
routes by specifying a new routes key entry.


stefanv:
See https://github.com/cesium-ml/baselayer/pull/263/files

PR #283, opened 2022-02-11
Reassign port to 5700, fix #282

### LINKED ISSUE #282: Mac OS Monterey uses port 5000 for Airplay
Skyportal uses port 5000 by default, but on the new Mac OS Monterey, this port is by default used for Airplay.

Some background: https://anandtripathi5.medium.com/port-5000-already-in-use-macos-monterey-issue-d86b02edd36c

You can manually disable Airplay receiver, in which case the port is freed up, but a better solution would be to use a different port. 5700 was a popular choice.

(Same issue as https://github.com/skyportal/skyportal/issues/2657)

robertdstein:
This PR changes the default baselayer port to 5700, from 5000. This avoids problems of port 5000 being occupied, as explained in #282.

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -90,7 +90,7 @@ log:
 ports:
     websocket: 64000
     fake_oauth: 63000
-    app: 5000
+    app: 5700
     app_http_proxy: 5001
     app_internal: 65000  # nginx forwards this port to ports:app
     dask: 63500


stefanv:
Thanks @robertdstein.

The port appears numerous times in the baselayer and SkyPortal repositories (maybe fritz too?), so please replace all instances and make associates PRs.

`rg` (ripgrep) is your friend!  If you don't have that tool (which is an awesome tool, btw), try `grep -r 5000`.

robertdstein:
So where exactly did we land on this? Is the idea still to simply transition the port number in situ, or to also gather all port references to one place?

stefanv:
I think we agreed that the easiest may be to advise mac users to add:

```
ports:
    app: 5050
```

(Or whatever the port should be) to their local `config.yaml`.

PR #284, opened 2022-02-11
RequestsSessionMixin

mcoughlin:
This PR uses RequestsSessionMixin, an updated class of Selenium Requests.


mcoughlin:
@stefanv It's the PR that let the SkyPortal tests pass!

PR #285, opened 2022-02-17
Boolean get_query_argument

guynir42:
We need to inherit `get_query_handler` to deal with situations where the expected output is a boolean, not a string. 

Example: `do_something = self.get_query_argument("doSomething", False)` followed by `if do_something:` will do the following:
- If the argument `doSomething` is not given in the query, return `False`, which is desired behavior. 
- If the argument is given as "True", returns the string `"True"` then the `if` will evaluate it as true. Desired behavior. 
- If the argument is given as "False" it returns the string `"False"` which the `if` takes as true. This is the wrong behavior. 

Instead, we override the `get_query_argument` function to act differently if given a default boolean. In that case, it will interpret any of the following inputs: `['true', 't', 'yes', '1']` (in a _case insensitive_ manner) as true. All other values are interpreted as false.


app/handlers/base.py
```diff
@@ -392,3 +392,9 @@ def push_notification(self, note, notification_type='info'):
             action='baselayer/SHOW_NOTIFICATION',
             payload={'note': note, 'type': notification_type},
         )
+
+    def get_query_argument(self, name, default=RequestHandler._ARG_DEFAULT, strip=True):
```
stefanv, line 396:
```suggestion
    def get_query_argument(self, *args, **kwargs):
```

app/handlers/base.py
```diff
@@ -392,3 +392,9 @@ def push_notification(self, note, notification_type='info'):
...
+        ret = RequestHandler.get_query_argument(self, name, default, strip)
```
stefanv, line 397:
```suggestion
        arg = super().get_query_argument(*args, **kwargs)
```

app/handlers/base.py
```diff
@@ -392,3 +392,9 @@ def push_notification(self, note, notification_type='info'):
...
+        if type(default) == bool:
```
stefanv, line 398:
```suggestion
        if type(kwargs.get('default', None)) == bool:
```

app/handlers/base.py
```diff
@@ -392,3 +392,9 @@ def push_notification(self, note, notification_type='info'):
...
+            ret = type(ret) == str and ret.lower() in ['true', 'yes', 't', '1']
```
stefanv, line 399:
```suggestion
            return arg.lower() in ['true', 'yes', 't', '1']
```

app/handlers/base.py
```diff
@@ -392,3 +392,9 @@ def push_notification(self, note, notification_type='info'):
...
+        return ret
```
stefanv, line 400:
```suggestion
```

app/handlers/base.py
guynir42: What if `arg` is None?

app/handlers/base.py
guynir42: I think we need to return after the `if` or we won't get the arguments for the non-boolean case.

app/handlers/base.py
stefanv: Yes, that can happen. How about `str(arg).lower() in ...`.

guynir42:
@stefanv ready for re-review.

stefanv:
It's a bit scary that we don't have tests for baselayer :/  Fortunately, it's well tested throug SkyPortal.
Thanks @guy.

PR #293, opened 2022-03-02
Use select instead of query for sqlalchemy

mcoughlin:
This PR begins to move the permissions infrastructure to use sqlalchemy's select instead of query.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -835,9 +835,10 @@ class BaseMixin:
             cls, user_or_token
         ).where(cls.id == self.id).subquery()
 
+        query = sa.select(sa.func.count(accessibility_table.columns.id))
         # Query for the value of the access_func for this particular record and
         # return the result.
-        result = DBSession().execute(sa.select(sa.func.count(accessibility_table.columns.id))).scalar_one() > 0
+        result = DBSession().execute(query).scalar_one() > 0
         if result is None:
             result = False
 
@@ -958,7 +959,8 @@ class BaseMixin:
             )
 
         logic = getattr(cls, mode)
-        query = logic.query_accessible_rows(cls, user_or_token, columns=columns)
+        query = logic.query_accessible_rows(cls,
+                                            user_or_token, columns=columns)
         for option in options:
             query = query.options(option)
         return query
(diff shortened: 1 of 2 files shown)


stefanv:
Hi Michael, thanks for this!

Since there are no baselayer tests, I'll have to rely on you testing this against SkyPortal (which I know you are doing).

Some questions:

1. I see sometimes select is used inside of session, sometimes not—how does that work?
2. Should we also move to the `with session` syntax?

Otherwise the changes look straightforward and we can merge this soon.  After that, I will enable black + pre-commit on this repo, so we don't have to worry about the formatting of all those long lines any more.

mcoughlin:
@stefanv 
So in baselayer, sometimes the requirement is to return just the query (and therefore just the select) and the results of the query (and therefore the session). That's why that is happening.

We can do the new syntax, but I thought that might be a second round for SP at least.

app/models.py
```diff
@@ -961,9 +959,11 @@ def query_records_accessible_by(
             )
 
         logic = getattr(cls, mode)
-        return logic.query_accessible_rows(
-            cls, user_or_token, columns=columns
-        ).options(options)
+        query = logic.query_accessible_rows(cls,
+                                            user_or_token, columns=columns)
+        for option in options:
+            query = query.options(option)
```
stefanv, line 965: Why the for loop instead of, say, `query.options(options)`?

stefanv (commented):
Thanks, Michael! This looks good; happy to merge.  I left one quick question.

app/models.py
mcoughlin: I found options wanted a single option and not a list when using select.

stefanv: But, but... https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.options

mcoughlin: I see, maybe ** instead of just a list then?

stefanv: Isn't options a list?

mcoughlin: [16:35:36 basehandler] Error response returned by [/api/internal/recent_gcn_events]: [ExecutionOption Core or ORM object expected, got [<sqlalchemy.orm.strategy_options._UnboundLoad object at 0x1676b3d00>].]

when passing it a list.

stefanv: 🤷‍♂️

PR #298, opened 2022-03-20
specify subquery loading for relationships

mcoughlin:
This PR explicitly sets subquery loading for relationships (e.g. https://docs.sqlalchemy.org/en/14/orm/loading_relationships.html)
which solves an issue where certain relationships aren't loaded by default.


stefanv:
@mcoughlin Can you update this PR and give a bit more context for what's happening here?

PR #312, opened 2022-05-03
add permissioned DBSession context manager

guynir42:


app/handlers/base.py
```diff
@@ -152,6 +153,53 @@ def bulk_verify(mode, collection, accessor):
...
+    def permissioned_session(self):
+        """Check that the current user has permission to create, read,
+        update, or delete rows that are present in the session. If not,
+        raise an AccessError (causing the transaction to fail and the API to
+        respond with 401).
+        If all permissions are satisfied, the session is committed.
+        All that, using a context manager. 
+        """
+        with DBSession() as session:
+            # get items to be inserted
```
stefanv, line 167: yield first

app/handlers/base.py
```diff
@@ -152,6 +153,33 @@ def bulk_verify(mode, collection, accessor):
 
 
 class BaseHandler(PSABaseHandler):
+    @contextmanager
+    def Session(self, use_auto_verify=True):
```
stefanv, line 157: Can we change this to just `verify`?

app/handlers/base.py
```diff
@@ -152,6 +153,33 @@ def bulk_verify(mode, collection, accessor):
...
+        """
+        Generate a scoped session that also has knowledge
+        of the current user, and that can automatically
+        verify and commit the changes to it when going out of context.
+        The current user is taken from the handler's current_user.
```
stefanv, line 162:
```suggestion
        The current user is taken from the handler's `current_user`.
```

app/handlers/base.py
```diff
@@ -152,6 +153,33 @@ def bulk_verify(mode, collection, accessor):
...
+        This is a shortcut method to models.Session
```
stefanv, line 163:
```suggestion
        This is a shortcut method to `models.Session`
```

app/handlers/base.py
```diff
@@ -152,6 +153,33 @@ def bulk_verify(mode, collection, accessor):
...
+        that saves the need to manually input the user object.
+
+        Parameters
+        ----------
+        use_auto_verify: boolean
+            if True (default), will call verify and commit
+            functions of the session before exiting the context.
```
stefanv, line 170:
```suggestion
            objects in the session before exiting the context.
```

stefanv: I think the behavior of this should be to verify *upon* commit, i.e. to override the commit of the returned Session.

app/handlers/base.py
```diff
@@ -152,6 +153,33 @@ def bulk_verify(mode, collection, accessor):
...
+
+        Returns
+        -------
+        a scoped session object that can be used in a context
```
stefanv, line 174:
```suggestion
        A scoped session object that can be used in a context
```

app/models.py
```diff
@@ -20,14 +22,112 @@
...
+    a scoped session object that can be used in a context
+    manager to access the database. If auto verify is enabled,
+    will use the current user given to apply verification
+    and commit to the database when exiting context.
+
+    """
+
+    def verify(self):
+        # import here to avoid circular import loop
+        from .handlers.base import bulk_verify
```
stefanv, line 65: Why not move `bulk_verify` to a more sensible place; it does not need to live on the handler.

app/models.py
```diff
@@ -20,14 +22,112 @@
...
+    if user_or_token is not None and not isinstance(user_or_token, (User, Token)):
+        raise ValueError(
+            "user_or_token must be an instance of User or Token, "
+            f"got {user_or_token.__class__.__name__}."
+        )
+    with scoped_session(sessionmaker(), scopefunc=session_context_id.get)() as session:
+        session.user_or_token = user_or_token
+        session.use_auto_verify = use_auto_verify
+        # make this an instance method of session
+        session.verify = types.MethodType(verify, session)
```
stefanv, line 118: Why not just assign the function?

app/models.py
```diff
@@ -20,14 +22,112 @@
...
+        yield session
+
+        session.expunge_all()  # make sure objects persist after session closes
```
stefanv, line 121: Why is this necessary? Do we want objects to persist after the session closes?

app/models.py
```diff
@@ -20,14 +22,112 @@
...
+
+        # this gets executed when external context is finished
+        if use_auto_verify:
```
stefanv, line 124: As mentioned elsewhere, this may be better attached to `session.commit`.

app/models.py
```diff
@@ -210,6 +308,27 @@ def query_accessible_rows(self, cls, user_or_token, columns=None):
 
         raise NotImplementedError
 
+    def select_accessible_rows(self, cls, user_or_token, columns=None):
+        """Construct a Select object that, when executed, returns the rows of a
+        specified table that are accessible to a specified user or token.
+
+        Parameters
+        ----------
+        cls: `baselayer.app.models.DeclarativeMeta`
```
stefanv, line 317: Throughout, note that the `:` requires a space before and after.

app/models.py
```diff
@@ -284,7 +403,6 @@ class Public(UserAccessControl):
     def query_accessible_rows(self, cls, user_or_token, columns=None):
```
stefanv, line 403: Do you want to add a deprecation warning or notice in the docs for the query functions?

app/models.py
```diff
@@ -948,13 +1327,142 @@ def query_records_accessible_by(
...
+        id_or_list,
+        user_or_token,
+        mode="read",
+        raise_if_none=False,
+        options=[],
+    ):
+        """Return a database record if it is accessible to the specified User or
+        Token. If no record exists, return None. If the record exists but is
+        inaccessible, raise an `AccessError`.
+        Parameters
```
stefanv, line 1364: Newline

app/models.py
```diff
@@ -948,13 +1327,142 @@ def query_records_accessible_by(
...
+                if options:
+                    stmt = sa.select(cls).options(options).where(cls.id == pk.item())
+                else:
+                    stmt = sa.select(cls).where(cls.id == pk.item())
+                instance = session.scalars(stmt).first()
+                if raise_if_none:
+                    if instance is None or not instance.is_accessible_by(
+                        user_or_token, mode=mode
+                    ):
+                        raise AccessError(f"Cannot find {cls.__name__} with id: {pk}")
```
stefanv, line 1398: Why do we raise here instead of just returning None?

app/models.py
```diff
@@ -1060,6 +1569,8 @@ def is_readable_by(self, user_or_token):
     def create_or_get(cls, id):
         """Return a new `cls` if an instance with the specified primary key
         does not exist, else return the existing instance."""
+        # with DBSession() as session:
```
stefanv, line 1572: Commented code

stefanv (commented):
Thanks @guynir42, this is a big chunk of work.

I wish we had some tests, but I think for that I'll have to first add sqlite as a DB option.

app/handlers/base.py
guynir42: I called the function `verify`. I can do `auto_verify` if that's better?

app/models.py
guynir42: You may be right. I was just trying not to change more than I had to. I will give it a try.

app/handlers/base.py
stefanv: I don't know the advantage of making the keyword longer. `verify` as a keyword seems to indicate perfectly well what is going to happen.

I don't think we should "auto" verify---just verify on request.

guynir42: So I can make the input name `verify` and also the function on the session called `verify`. I will just rename the function definition to something like `run_verification` which is never exposed to the user, so it should be ok.

stefanv: But why not also expose `verify` to the user?

guynir42: Yes, I will still name it `session.verify` and the control boolean will be named `session.use_auto_verify` but when inputting the function the parameter will be called just `verify`.

stefanv: Ah! For storing internal state, it is customary to prefix with an underscore: `session._verify`.

guynir42: That's fine in general, but I feel that would be confusing as there will be a method called `verify`. Let me push another commit and we'll see if it looks ok.

app/handlers/base.py
guynir42: Not sure what you mean...?

app/models.py
guynir42: Because I want the function to take the `session` as the `self` argument.

app/models.py
guynir42: Yes. When committing the session, objects we got from it become stale. There are lots of cases where you keep using them, even when they are detached from the DB. Example: use a `with Session() as session` inside a function that returns an object you need. The function quits and shuts down the session but the object is still usable.

app/models.py
guynir42: I've changed the access pattern here. It makes sure that if the record doesn't exist OR if it inaccessible the results are exactly the same.

app/handlers/base.py
guynir42: You mean that every time `commit()` is called it will also call `verify()`? That means we will also need to have a switch controlling whether we want to automatically commit at end of session, on top of the one that makes it automatically verify before each commit.

app/models.py
guynir42: yeah I will add something to the doc strings.

guynir42:
@stefanv this is ready for another review.

app/handlers/base.py
stefanv: I don't think we want to modify the behavior of existing sessions, other than to add verification. Sessions do not automatically commit, so it doesn't make sense to automatically verify.  Commit should trigger a verify for our special session.

guynir42: I think it could be nice to produce a session that also commits and verifies when exiting. That's an optional thing, so we can have it default to False, but still allow the user to start a session, in context, that would (verify and) commit at exit.

stefanv: I think a nice-to-have does not outweigh the inconsistency this will introduce with SQLAlchemy.  The risk of introducing programming errors because users are looking at the SQLAlchemy docs is too big, especially since it is done silently.

And we don't really need it, do we? It's mainly a convenience for not typing `session.commit()` when you're ready to write to the DB?

stefanv: I think having `session.commit()` verify, on the other hand, is a perfectly reasonable addition.

guynir42: I guess that's true. I was just thinking that while we are adding stuff at the end of the context, might as well do that also. But I will remove it.

app/models.py
guynir42: If we do not commit the session we probably don't need the expunge anymore. But is it ok to keep using an object loaded after a session that was closed (but not committed)?

stefanv: I don't think you can use objects retrieved outside of the session they came from.

guynir42: Yeah, you must. For example the `get_if_accessible_by` function returns an object. There's no way to recreate this pattern if you create a session and want to pass the object out in the return statement. 
Besides, every time a handler gives back data, the session ends and the data persists. 

I think the only thing that is not allowed is to modify the objects after the session is closed, which makes sense.

stefanv: That's correct—no writing to those objects.

guynir42: OK. I've also removed the `expunge_all()` now that we are not committing automatically. 
That means that commit is only used when making changes to the DB. 
When only getting information, we must rely on the select statements that go through baselayer to give us only accessible rows, and not the `verify()` command.

app/handlers/base.py
```diff
@@ -98,60 +97,40 @@ def log_exception(self, typ=None, value=None, tb=None):
...
 class BaseHandler(PSABaseHandler):
+    @contextmanager
+    def Session(self, verify=True):
+        """
+        Generate a scoped session that also has knowledge
+        of the current user, and that can automatically
+        verify and commit the changes to it when going out of context.
+        The current user is taken from the handler's `current_user`.
+        This is a shortcut method to `models.Session`
+        that saves the need to manually input the user object.
```
stefanv, line 114: How would you input the user object? Don't you need a Session to get it?

app/handlers/base.py
```diff
@@ -98,60 +97,40 @@ def log_exception(self, typ=None, value=None, tb=None):
...
+
+        Parameters
+        ----------
+        verify : boolean
```
stefanv, line 118: Needs update; verify will only execute on commit

app/handlers/base.py
```diff
@@ -98,60 +97,40 @@ def log_exception(self, typ=None, value=None, tb=None):
...
+            if True (default), will call the functions
+            `verify()` and `commit()` of the session
+            before exiting the context.
+
+        Returns
+        -------
+        A scoped session object that can be used in a context
+        manager to access the database. If auto verify is enabled,
```
stefanv, line 126: Also here

guynir42:
@stefanv can we move forward with this?

app/handlers/base.py
```diff
@@ -55,7 +54,9 @@ def get_current_user(self):
         user_id = self.user_id()
         oauth_uid = self.get_secure_cookie("user_oauth_uid")
         if user_id and oauth_uid:
-            user = User.query.get(int(user_id))
+            # user = User.query.get(int(user_id))
+            with Session(user=None, verify=False) as session:
```
stefanv, line 58: I suppose `user` is None by default, unless using `self.Session`?

app/handlers/base.py
```diff
@@ -98,60 +99,40 @@ def log_exception(self, typ=None, value=None, tb=None):
...
+        are accessible to the user.
+        The current user is taken from the handler's `current_user`.
+        This is a shortcut method to `models.Session`
+        that saves the need to manually input the user object.
+
+        Parameters
+        ----------
+        verify : boolean
+            if True (default), will call the functions
+            `verify()` and whenever `commit()` is called.
```
stefanv, line 123: Typo?

app/models.py
```diff
@@ -20,14 +23,167 @@
...
+        If True (default), will call the verify function
+        of the session before each call to commit().
+        If True, must specify a legal User object.
+
+    Returns
+    -------
+    a scoped session object that can be used in a context
+    manager to access the database. If verify is enabled,
+    the session will also verify the user has access
+    to all rows before any commit() command.
```
stefanv, line 60: whenever the commit method is called

app/models.py
```diff
@@ -20,14 +23,167 @@
...
+    ----------
+    user_or_token : baselayer.app.models.User object
+        or baselayer.app.models.Token object.
+        The object representing the current user.
+        Can be None only if the verify parameter
+        is set to False.
+    verify : boolean
+        If True (default), will call the verify function
+        of the session before each call to commit().
+        If True, must specify a legal User object.
```
stefanv, line 53: legal -> valid

app/models.py
```diff
@@ -20,14 +23,167 @@
...
+        ):
+            bulk_verify(mode, collection, self.user_or_token)
+
+        # update transaction state in DB, but don't commit yet. this updates
+        # or adds rows in the database and uses their new state in joins,
+        # for permissions checking purposes.
+        self.flush()
+        bulk_verify("create", new_rows, self.user_or_token)
+
+    def run_commit(self):
```
stefanv, line 103: Why not just override `self.commit`?

app/models.py
```diff
@@ -20,14 +23,167 @@
...
 use_webhook = cfg["security.slack.enabled"]
 webhook_url = cfg["security.slack.url"]
 
+session_context_id = contextvars.ContextVar("request_id", default=None)
+# left here for backward compatibility:
+DBSession = scoped_session(sessionmaker(), scopefunc=session_context_id.get)
+
+
+@contextmanager
+def Session(user_or_token=None, verify=True):
```
stefanv, line 37: Why is Session a function and not a subclass?

stefanv (commented):
Biggest issue remaining: Session should probably subclass from SQLAlchemy Session instead of being a function.

PR #316, opened 2022-05-19
Select statements and Session

guynir42:
his PR introduces scoped sessions with user access verification.

It also introduces new methods on the Base model, to allow querying using the new "select" style compatible with SQLA 2.0

It also fixes the behavior of `get_if_accessible_by` to respond exactly the same if the row requested does not exist or if it is not accessible. 

This is mostly backward compatible, but a separate PR to skyportal fixes the few issues in the handlers and the tests that are broken by this change. 

Most importantly, it allows us to move each handler separately into the new SQLA 2.0 style without having to change everything at once.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -302,8 +302,9 @@ class Public(UserAccessControl):
         """
         # return only selected columns if requested
         if columns is not None:
-            return DBSession().query(*columns).select_from(cls)
-        return DBSession().query(cls)
+            return sa.select(*columns).select_from(cls)
+        else:
+            return sa.select(cls)
 
 
 public = Public()
@@ -369,9 +370,9 @@ class AccessibleIfUserMatches(UserAccessControl):
 
         # return only selected columns if requested
         if columns is not None:
-            query = DBSession().query(*columns).select_from(cls)
+            query = sa.select(*columns).select_from(cls)
         else:
-            query = DBSession().query(cls)
+            query = sa.select(cls)
 
         # traverse the relationship chain via sequential JOINs
         for relationship_name in self.relationship_names:
@@ -386,7 +387,7 @@ class AccessibleIfUserMatches(UserAccessControl):
 
         # filter for records with at least one matching user
         user_id = self.user_id_from_user_or_token(user_or_token)
-        query = query.filter(cls.id == user_id)
+        query = query.where(cls.id == user_id)
         return query
 
     @property
@@ -487,9 +488,9 @@ class AccessibleIfRelatedRowsAreAccessible(UserAccessControl):
 
         # return only selected columns if requested
         if columns is None:
-            base = DBSession().query(cls)
+            base = sa.select(cls)
         else:
-            base = DBSession().query(*columns).select_from(cls)
+            base = sa.select(*columns).select_from(cls)
 
         # ensure the target class has all the relationships referred to
         # in this instance
@@ -609,7 +610,7 @@ class ComposedAccessControl(UserAccessControl):
 
         # retrieve specified columns if requested
         if columns is not None:
-            query = DBSession().query(*columns).select_from(cls)
+            query = sa.select(*columns).select_from(cls)
         else:
             query = DBSession().query(cls)
 
@@ -654,7 +655,7 @@ class ComposedAccessControl(UserAccessControl):
         # in the case of or logic, require that only one of the conditions be
         # met for each row
         if self.logic == "or":
-            query = query.filter(
+            query = query.where(
                 sa.or_(*[col.isnot(None) for col in accessible_id_cols])
             )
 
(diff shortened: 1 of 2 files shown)


app/models.py
```diff
@@ -20,14 +21,155 @@
...
+    Create an instance of Session when you
+    want to apply a verification method on all added
+    or modified or deleted rows before commiting them
+    to the database.
+
+    This class overrides the commit() function
+    by adding a verify() function before it.
+
+    Use this in a context manager:
+    with VerifiedSession(user_object) as session:
```
stefanv, line 45: Are you still planning to rename to VerifiedSession as indicated here?

app/models.py
```diff
@@ -20,14 +21,155 @@
...
+        ...
+        session.commit()
+
+    This will make sure the changes to the DB
+    are verified, and will close the connection
+    when the context goes out.
```
stefanv, line 51:
```suggestion
    when exiting the context.
```

app/models.py
```diff
@@ -20,14 +21,155 @@
...
+            ).first()
+
+        Parameters
+        ----------
+        user_or_token: baselayer.app.models.User object
+            or baselayer.app.models.Token object.
+            The object representing the current user.
+
+        """
+        self.user_or_token = user_or_token
```
stefanv, line 74: Remember to also call `super().__init__()`

app/models.py
```diff
@@ -20,14 +21,155 @@
...
+        self.flush()
+        bulk_verify("create", new_rows, self.user_or_token)
+
+    def commit(self):
+        print("verifying rows before committing... ")
+        self.verify()
+        super().commit()
+
+
+VerifiedSession = scoped_session(
```
stefanv, line 121: It's confusing to have Session and VerifiedSession being entirely different things. Maybe the class should be called `_VerifiedSession` to make clear that `VerifiedSession` is the one to use?

app/models.py
```diff
@@ -305,6 +471,30 @@ def query_accessible_rows(self, cls, user_or_token, columns=None):
             return DBSession().query(*columns).select_from(cls)
         return DBSession().query(cls)
 
+    def select_accessible_rows(self, cls, user_or_token, columns=None):
+        """Construct a Select object that, when executed, returns the rows of a
+        specified table that are accessible to a specified user or token.
+
+        Parameters
+        ----------
+        cls: `baselayer.app.models.DeclarativeMeta`
```
stefanv, line 480:
```suggestion
        cls : `baselayer.app.models.DeclarativeMeta`
```

app/models.py
```diff
@@ -305,6 +471,30 @@ def query_accessible_rows(self, cls, user_or_token, columns=None):
...
+            The mapped class of the target table.
+        user_or_token: `baselayer.app.models.User` or `baselayer.app.models.Token`
```
stefanv, line 482:
```suggestion
        user_or_token : `baselayer.app.models.User` or `baselayer.app.models.Token`
```

app/models.py
```diff
@@ -305,6 +471,30 @@ def query_accessible_rows(self, cls, user_or_token, columns=None):
...
+            The User or Token to check.
+        columns: list of sqlalchemy.Column, optional, default None
```
stefanv, line 484:
```suggestion
        columns : list of sqlalchemy.Column, optional, default None
```

app/models.py
```diff
@@ -389,6 +581,51 @@ def query_accessible_rows(self, cls, user_or_token, columns=None):
         query = query.filter(cls.id == user_id)
         return query
 
+    def select_accessible_rows(self, cls, user_or_token, columns=None):
+        """Construct a Select object that, when executed, returns the rows of a
+        specified table that are accessible to a specified user or token.
+
+        Parameters
+        ----------
+        cls: `baselayer.app.models.DeclarativeMeta`
```
stefanv, line 590:
```suggestion
        cls : `baselayer.app.models.DeclarativeMeta`
```

app/models.py
```diff
@@ -389,6 +581,51 @@ def query_accessible_rows(self, cls, user_or_token, columns=None):
...
+            The mapped class of the target table.
+        user_or_token: `baselayer.app.models.User` or `baselayer.app.models.Token`
```
stefanv, line 592:
```suggestion
        user_or_token : `baselayer.app.models.User` or `baselayer.app.models.Token`
```

stefanv (commented):
OK, great, I think this is just about there.  Let's just clarify Session vs VerifiedSession then we can get this in. Thanks!

app/models.py
guynir42: It is renamed in the function call below the class definition.

app/models.py
guynir42: Good call.

PR #318, opened 2022-05-26
More upgrades to session

guynir42:
While running tests on skyportal with the new SQLA 2.0 tools, I ran into some more bugs. This tries to fix those, together with running the entire SP test suite (in a separate PR for SP).

### DIFF
--- a/app/access.py
+++ b/app/access.py
@@ -1,9 +1,11 @@
 import functools
 
+import sqlalchemy as sa
 import tornado.web
+from sqlalchemy.orm import joinedload
 
 from baselayer.app.custom_exceptions import AccessError  # noqa: F401
-from baselayer.app.models import Role, Token, User  # noqa: F401
+from baselayer.app.models import DBSession, Role, Token, User  # noqa: F401
 
 
 def auth_or_token(method):
@@ -24,7 +26,17 @@ def auth_or_token(method):
         token_header = self.request.headers.get("Authorization", None)
         if token_header is not None and token_header.startswith("token "):
             token_id = token_header.replace("token", "").strip()
-            token = Token.query.get(token_id)
+            with DBSession() as session:
+                token = session.scalars(
+                    sa.select(Token)
+                    .options(
+                        joinedload(Token.created_by).options(
+                            joinedload(User.acls),
+                            joinedload(User.roles),
+                        )
+                    )
+                    .where(Token.id == token_id)
+                ).first()
             if token is not None:
                 self.current_user = token
                 if not token.created_by.is_active():
--- a/app/handlers/base.py
+++ b/app/handlers/base.py
@@ -140,6 +140,8 @@ class BaseHandler(PSABaseHandler):
 
         """
         with VerifiedSession(self.current_user) as session:
+            # must merge the user object with the current session
+            # ref: https://docs.sqlalchemy.org/en/14/orm/session_basics.html#adding-new-or-existing-items
             session.add(self.current_user)
             yield session
 
--- a/app/handlers/base.py
+++ b/app/handlers/base.py
@@ -14,7 +14,6 @@ import social_tornado.handlers as psa_handlers
 # be used to look up the logged in user.
 import sqlalchemy
 import tornado.escape
-from sqlalchemy.orm import joinedload
 from tornado.log import app_log
 from tornado.web import RequestHandler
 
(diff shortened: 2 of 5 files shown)


app/access.py
```diff
@@ -24,7 +26,17 @@ def wrapper(self, *args, **kwargs):
         token_header = self.request.headers.get("Authorization", None)
         if token_header is not None and token_header.startswith("token "):
             token_id = token_header.replace("token", "").strip()
-            token = Token.query.get(token_id)
+            with DBSession() as session:
+                token = session.scalars(
+                    sa.select(Token)
+                    .options(
+                        joinedload(Token.created_by).options(
```
stefanv, line 33: Any reason not to have these loaded by default every time the token is retrieved?

app/handlers/base.py
```diff
@@ -58,7 +59,9 @@ def get_current_user(self):
         if user_id and oauth_uid:
             with DBSession() as session:
                 user = session.scalars(
-                    sqlalchemy.select(User).where(User.id == user_id)
+                    sqlalchemy.select(User)
+                    .where(User.id == user_id)
+                    .options(joinedload(User.acls))
```
stefanv, line 64: Same here

app/handlers/base.py
```diff
@@ -136,7 +139,8 @@ def Session(self, verify=True):
         before every commit.
 
         """
-        with VerifiedSession(self.current_user, verify) as session:
+        with VerifiedSession(self.current_user) as session:
+            session.add(self.current_user)
```
stefanv, line 143: Did you ever figure out if this is the right way?

app/models.py
```diff
@@ -118,9 +118,11 @@ def commit(self):
         super().commit()
 
 
-VerifiedSession = scoped_session(
-    sessionmaker(class_=_VerifiedSession), scopefunc=session_context_id.get
-)
+def VerifiedSession(user_or_token):
```
stefanv, line 121: Ah, interesting. The session_maker doesn't do this by default?

app/handlers/base.py
guynir42: I'm not 100% sure. Let me check out the docs (I recall reading this but I need to make sure).

app/models.py
guynir42: Doesn't do what by default?

app/handlers/base.py
guynir42: Yeah, they say we should use `add()` https://docs.sqlalchemy.org/en/14/orm/session_basics.html#adding-new-or-existing-items

app/access.py
guynir42: Maybe we should replace all the default eager loading with `selectin` loads? 
https://docs.sqlalchemy.org/en/14/orm/loading_relationships.html#select-in-loading

doc/dev.md
```diff
@@ -87,19 +94,22 @@ The `get` and `get_all` functions open a session internally and retrieve the obj
 if they are accessible to the user. In the case of `get`, if any of the IDs given (as a scalar or list)
 are not accessible to do not exist in the DB, the function returns None, or raises an `AccessError`
 (if `raise_if_none=True` is specified). The `get_all` just retrieves all rows that are accessible from that table.
+Note that these two methods will produce an object _not associated with the external session, if any_.
```
stefanv, line 97: This should probably be clarified a bit.

app/models.py
```diff
@@ -1598,7 +1604,11 @@ def __repr__(self):
     def to_dict(self):
         """Serialize this object to a Python dictionary."""
         if sa.inspection.inspect(self).expired:
-            DBSession().refresh(self)
+            try:
+                self = DBSession().merge(self)
+                DBSession().refresh(self)
+            except sa.orm.exc.ObjectDeletedError:
```
stefanv, line 1610: Perhaps this error should just propagate.

app/access.py
stefanv: I suppose it doesn't matter much, but I like the simplicity of that statement.

stefanv (approved):

app/models.py
guynir42: I think I added this because in some cases the object is deleted but you don't want an error (like when tearing down the fixtures).

stefanv: This means returning different representations of the object at different times, which sounds like a recipe for disaster.

guynir42: Yeah that's a good point. Let me try it and see how disruptive it is.

guynir42: I mean how disruptive it is to let the exception propagate...

stefanv: Ah lol

app/access.py
guynir42: and we cannot load Tokens by default because it doesn't pass the permissions checks since user1 cannot see the token of user2.

guynir42:
@stefanv I fixed/answer all comments and the tests are :heavy_check_mark:

PR #319, opened 2022-05-26
Add  method for Config

stefanv:
@Theodlz This should make your life a bit simpler in your default roles PR.

Now you can do `cfg.get('user.default_roles', [])`

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -63,11 +63,18 @@ class Config(dict):
 
         val = self
         for key in keys:
-            val = val.get(key)
-            if val is None:
-                return None
+            if isinstance(val, dict):
+                val = dict.__getitem__(val, key)
+            else:
+                raise KeyError(key)
 
         return val
+
+    def get(self, key, default=None, /):
+        try:
+            return self.__getitem__(key)
+        except KeyError:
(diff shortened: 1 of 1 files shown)


Theodlz:
Very nice ! Thanks a lot.

app/config.py
```diff
@@ -63,12 +63,19 @@ def __getitem__(self, key):
...
-            if val is None:
-                return None
+            if isinstance(val, dict):
+                val = dict.__getitem__(val, key)
+            else:
+                raise KeyError(key)
 
         return val
 
+    def get(self, key, default=None, /):
```
profjsb, line 73: I don't understand the need for a slash at the end here.

stefanv: It is the same signature as what Python has for its dictionary; it means you cannot specify any arguments before the slash as keyword arguments.  I.e., you cannot write `get(key, default='foo')` but have to write `get(key, 'foo')`.  I'm just following the existing Python pattern here (personally, I don't really see the point).

profjsb: Thanks for that explanation

profjsb (approved):

PR #323, opened 2022-06-13
Added user affiliation

Theodlz:
This very short PR adds the user affiliation to the user model.
This field can be useful when publishing articles or writing circulars based on the science made by users.
A good example is writing GCN circulars on SkyPortal, where we will need the affiliation of users.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1809,6 +1809,7 @@ class User(Base):
 
     first_name = sa.Column(sa.String, nullable=True, doc="The User's first name.")
     last_name = sa.Column(sa.String, nullable=True, doc="The User's last name.")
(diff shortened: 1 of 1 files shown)


profjsb:
How should we handle multiple affiliations? And is this the full (LaTeX) affiliation string or free-form text? This could be a list of strings (default []) something like:

```
["Department of Mathematics, University X", "Department of Biology, University Y"]
```

Which gets rendered as:

\affil[1]{Department of Mathematics, University X}
\affil[2]{Department of Biology, University Y}

In LaTeX. Or as a comma-separated string otherwise. 

After this PR is merged we may want to render this affiliation in the user prefs.

PR #326, opened 2022-06-15
Add tool to blend logs, taking timestamps into consideration

stefanv:
@mcoughlin Here is another approach to addressing https://github.com/skyportal/skyportal/issues/3084

This does not currently support the nginx log, but we can add that.

### DIFF
--- a/tools/merge_logs.py
+++ b/tools/merge_logs.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python
+#
+# Blend log files, keeping events in order.
+#
+
+import sys
+import re
+from datetime import datetime
+
+
+if len(sys.argv) < 2:
+    print("Usage: log_parse.py filename0.log filename1.log ...")
+    sys.exit(1)
+
+
+# 7-bit C1 ANSI sequences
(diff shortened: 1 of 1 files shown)


mcoughlin:
@stefanv Oh wow that's nifty. Maybe it could optionally return the log string if called as a function for potential parsing and display of the health events on a dedicated page like I was thinking about in that other PR?

stefanv:
Is there a specific reason we need to expose the logs via the web?

mcoughlin:
@stefanv i am thinking it would be very nice to at a glance (in an admin only page) to be able to see the logging rather than needing to be logged in and pulling down logs and such. It's mostly about convenience of rapid assessment.

stefanv:
Okay, maybe we should start with a tool and then talk about expanding scope. I'm a bit nervous, from a security perspective, to have log files served up. Perhaps it's fine, but then we may also want to consider whether we don't just want to use Google or GrayLog or similar.

mcoughlin:
@stefanv sounds good. I am cool with your tool for sure.

PR #327, opened 2022-06-15
Patch config get

guynir42:
The new changes in `app/config.py` make the behavior of the config more consistent with python dictionaries:
- when using `cfg['some_key']` it raises a `KeyError` if not found. 
- using `cfg.get('some_key')` will return `None` or whatever default was specified.
This is a good thing to have, but requires changing many, many files in skyportal. 
I suggest this patch for now, which still allows using `cfg.get()` with a default (or return `None`) but leaves the `gcf['...']` as it was before. 
An accompanying PR to revert these changes can be made after all the relevant places in skyportal are migrated to using `cfg.get()`.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -60,18 +60,25 @@ class Config(dict):
 
     def __getitem__(self, key):
         keys = key.split(".")
-
-        val = self
-        for key in keys:
-            if isinstance(val, dict):
-                val = dict.__getitem__(val, key)
-            else:
-                # raise KeyError(key)
-                return None
-        return val
+        try:
+            val = self
+            for key in keys:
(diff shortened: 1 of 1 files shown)


stefanv:
We should only need to patch skyportal in instances where we check the returned config value to be none. Are there really that many?

guynir42:
@stefanv there's quite a bit of code where the `cfg[...]` is used. In many places there's an assumption that it returns None. 
There's more than "a few" places, and I think the fix to that should come in a separate PR.

guynir42:
OK fine, I will add that.

stefanv:
Please see https://github.com/skyportal/skyportal/pull/3091

PR #328, opened 2022-06-16
User model - Affiliations

Theodlz:
This PR adds affiliations to the user model. The affiliations are stored as a list of strings, which can be displayed on the frontend and used to write scientific papers or circulars for example. 

A WIP PR on SkyPortal : https://github.com/skyportal/skyportal/pull/3092 uses this to display it on the frontend.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1809,6 +1809,7 @@ class User(Base):
 
     first_name = sa.Column(sa.String, nullable=True, doc="The User's first name.")
     last_name = sa.Column(sa.String, nullable=True, doc="The User's last name.")
+    affiliation = sa.Column(sa.String, nullable=True, doc="The User's affiliation.")
     contact_email = sa.Column(
         EmailType(),
         nullable=True,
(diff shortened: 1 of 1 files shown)


Theodlz:
Hi @profjsb. I renamed the branch affiliations just to make it more clear but that closed the previous PR. I added your suggestion and opened a WIP PR on SkyPortal.

guynir42 (approved):
Maybe add an example affiliation in the doc string?

Theodlz:
By the way, do you think I could have the permissions to run the GA ?

guynir42:
@stefan can probably add your permissions to the repo.

guynir42:
Note that you will need to add a PR on skyportal with a migration script (and I assume some code to display/edit affiliations somewhere).

guynir42:
> Note that you will need to add a PR on skyportal with a migration script (and I assume some code to display/edit affiliations somewhere).

Sorry I noticed there's already a skyportal PR as well.

guynir42:
@mcoughlin @Theodlz sure but I'm not sure what we need to coordinate. 
If this is ready (and passes skyportal tests) then we should merge it and then have your PR pin this version of baselayer. 
If that is ready then there's no reason not to merge. 
I would be extra careful to make sure the skyportal version correctly migrates the changes in baselayer!

mcoughlin:
@guynir42 ah by coordinate I just meant exactly that, I.e it will require that the SP PR go in so that the migration is in. If it was to get pinned without the migration we would be in trouble.

PR #329, opened 2022-06-20
allow longer for app to boot up in  `test_frontend `

profjsb:
Allow the app longer time to come up. Also wait until supervisord.conf is built before polling supervisor status (avoids complaining about conf not found)

### DIFF
--- a/tools/test_frontend.py
+++ b/tools/test_frontend.py
@@ -46,11 +46,14 @@ def all_services_running():
     return running if return_code in (0, 3) else False
 
 
-def verify_server_availability(url, timeout=60):
+def verify_server_availability(url, timeout=180):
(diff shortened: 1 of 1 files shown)


tools/test_frontend.py
```diff
@@ -46,11 +46,14 @@ def all_services_running():
     return running if return_code in (0, 3) else False
 
 
-def verify_server_availability(url, timeout=60):
+def verify_server_availability(url, timeout=180):
     """Raise exception if webservices fail to launch or connection to `url` is not
     available.
     """
     for i in range(timeout):
+        if not os.path.exists("baselayer/conf/supervisor/supervisor.conf"):
```
profjsb, line 54: this is the file location where `setup_services` writes it's unified supervisord.conf file.

PR #331, opened 2022-07-07
Try increasing nginx timeout a bit

mcoughlin:
This PR increases the nginx timeout to try to improve app responsivity.


mcoughlin:
@stefanv and I had a conversation on slack where he suggests we also add some sort of specific header to the frontend API calls so the app can know to prioritize those. But that feels like a development that will take a few weeks and this seems worth trying anyways to find out if tuning this number might help user experience anyways.

PR #332, opened 2022-07-14
Make sure cron.py gets a valid jobs list

profjsb:


stefanv:
This patch takes it from new style to old style.

profjsb:
Look at any cron.log of recent test failures on SP--you'll why we need this fix. (I uncovered it locally when the app failed to fire for me). We haven't had a full test suite pass on main for awhile because of this bug.

profjsb:
cfg.get only shows up 1 other place in baselayer:
```
$ rg "cfg.get"
services/external_logging/external_logging.py
59:    service_configs = cfg.get("external_logging", [])
```

profjsb:
> @profjsb I am not sure why `get` fails here. It is supposed to do exactly what you want. Are you sure this is the only place in the code where this key is loaded?

I dont know why it's failing but it is

guynir42:
@profjsb check that you are on the latest version of both skyportal and baselayer...

profjsb:
See e.g. https://github.com/skyportal/skyportal/actions/runs/2667271046

guynir42:
I think I know what is going on. 
We have `cron` setup in baselayer but then cancel it in higher config files by leaving that key empty. 
That sends a `None` instead of missing key (which `get` would have successfully turned into `[]`).
The solution is not to get rid of `get` but to test for `None` explicitly.

profjsb:
> I think I know what is going on. We have `cron` setup in baselayer but then cancel it in higher config files by leaving that key empty. That sends a `None` instead of missing key (which `get` would have successfully turned into `[]`). The solution is not to get rid of `get` but to test for `None` explicitly.

this is what my fix does -- it makes sure that `jobs` is a list which the remainder of the code requires. 

Note this problem has been happening for at least weeks (I just checked some logs on github for random commits). It causes the "model logic" test to fail a lot.

services/cron/cron.py
```diff
@@ -14,7 +14,7 @@
 log = make_log("cron")
 
 env, cfg = load_env()
-jobs = cfg.get("cron", [])
+jobs = cfg["cron"] or []
```
guynir42, line 17:
```suggestion
jobs = cfg.get("cron") or []
```

guynir42: This way you are catching both an empty key and a missing key.

guynir42:
@profjsb yeah it looks good. Go ahead and merge+pin.

guynir42:
> This is fine; the easy fix would have been to fix the config file that set the cron value to None instead of to `[]` (as it should)`.

I told you the user will always find a way to screw up the config file in ways you didn't anticipate. Even when the user is us.

PR #333, opened 2022-07-14
Have status_server return 200 when starting

mcoughlin:
This PR modifies status_server to return 200's instead of 503's based on the GCP warnings: "Unhealthy
Readiness probe failed: HTTP probe failed with statuscode: 503".


profjsb (approved):
Let's try it!

PR #334, opened 2022-07-21
Extend timeout some more to see if it makes a difference on production

mcoughlin:
This PR extends the nginx timeout further to see if it improves responsivity (or not).


profjsb (approved):
Happy to try this. Might check on preview if this has any impact on frontend experiences.

profjsb (approved):
Happy to try this. Might check on preview if this has any impact on frontend experiences.

PR #335, opened 2022-08-11
Allow overlaps in join_model

guynir42:
This PR is part of an effort to silence (but not fix) the SQLA warnings we have on relationships. 

I've allowed passing the `overlaps` argument to be added to each relationship that is defined inside `app.models.join_model`. 
This makes it possible to silence warnings, although a more serious investigation should be made into the root cause of the warnings. 

I've also allowed passing `new_name` as an argument that allows defining the relationship model's name, instead of the default which is just the mashup of the two model names. This removes warnings where multiple mappers have the same name. 

Finally, I've also made the output of `app.model_util.create_tables` be more compact. 

This, in combination with a skyportal PR will make the outputs much shorter.

### DIFF
--- a/app/model_util.py
+++ b/app/model_util.py
@@ -57,9 +57,10 @@ def create_tables(retry=5, add=True):
             print(f"Creating tables on database {conn.url.database}")
             models.Base.metadata.create_all()
 
-            print("Refreshed tables:")
-            for m in models.Base.metadata.tables:
-                print(f" - {m}")
+            table_list = ", ".join(list(models.Base.metadata.tables.keys()))
+            print(f"Refreshed tables: {table_list}")
+            # for m in models.Base.metadata.tables:
+            #     print(f" - {m}")
 
             return
 
--- a/app/models.py
+++ b/app/models.py
@@ -1682,6 +1682,8 @@ def join_model(
     fk_1="id",
     fk_2="id",
     base=Base,
+    new_name=None,
+    overlaps=None,
 ):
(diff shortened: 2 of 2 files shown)


app/model_util.py
```diff
@@ -57,9 +57,10 @@ def create_tables(retry=5, add=True):
             print(f"Creating tables on database {conn.url.database}")
             models.Base.metadata.create_all()
 
-            print("Refreshed tables:")
-            for m in models.Base.metadata.tables:
-                print(f" - {m}")
+            table_list = ", ".join(list(models.Base.metadata.tables.keys()))
+            print(f"Refreshed tables: {table_list}")
+            # for m in models.Base.metadata.tables:
```
mcoughlin, line 62: Do you mean to leave the print statements in?

app/models.py
```diff
@@ -1737,17 +1746,28 @@ def join_model(
         ),
     }
 
+    if overlaps:
+        if isinstance(overlaps, str):
+            overlap_string = overlaps
+        else:
+            overlap_string = ", ".join(overlaps)
```
mcoughlin, line 1753: should you explicitly check for a list?

mcoughlin (approved):
Just want to check maybe for being passed a list.

app/model_util.py
guynir42: Yeah. I'm replacing one print statement that is pretty long, with a more condensed one.

app/models.py
guynir42: Yes!

PR #339, opened 2022-08-24
Revert changes in status_server

guynir42:
Revert the status code 200 back to 503 in `status_server`. 
This solves the problem that some of the tests cannot start on CI.


mcoughlin:
@guynir42 I think we need to discuss this as this also brings down GCP.

mcoughlin (approved):
LGTM

PR #340, opened 2022-09-05
Bump node to 18 and fix GHA

profjsb:
Bumps node to 18 instead of 17. Also fixes the GHA to always test against the baselayer app. Also, splits out the linter into a separate test.


guynir42 (approved):
LGTM

PR #341, opened 2022-09-20
Add request_length to nginx log

mcoughlin:
This PR adds request_length to nginx log, which should allow us to track query sizes by IP. See e.g. https://www.ruby-forum.com/t/logging-variables-bytes-sent-where-is-bytes-read/241597/8.


stefanv:
Let's add something like "rl=" to the line to distinguish this field from the one next to it? Otherwise lgtm.

mcoughlin:
@stefanv Done!

PR #342, opened 2022-09-27
cron: do not fail if a job entry is empty

profjsb:
skips over jobs that are in the config but missing key parameters.


guynir42 (approved):
LGTM

PR #344, opened 2022-10-19
Return 404 rather than throw app error

mcoughlin:
This PR returns 404 rather than throw app error in case of bad query.

See https://github.com/skyportal/skyportal/issues/3603

### DIFF
--- a/app/access.py
+++ b/app/access.py
@@ -72,7 +72,14 @@ def permissions(acl_list):
                 or "System admin" in self.current_user.permissions
             ):
                 raise tornado.web.HTTPError(401)
-            return method(self, *args, **kwargs)
+            print(method, args, kwargs)
+            try:
+                return method(self, *args, **kwargs)
+            except TypeError as e:
+                raise tornado.web.HTTPError(
+                    404,
+                    f"Query exception occured: {str(e)}",
+                )
 
         return wrapper
 


app/access.py
```diff
@@ -72,7 +72,14 @@ def wrapper(self, *args, **kwargs):
                 or "System admin" in self.current_user.permissions
             ):
                 raise tornado.web.HTTPError(401)
-            return method(self, *args, **kwargs)
+            print(method, args, kwargs)
+            try:
+                return method(self, *args, **kwargs)
+            except TypeError as e:
```
stefanv, line 78: Under what circumstance does the method return a TypeError?

I think this is in response to the request you had to 404 when `comment` instead of `comments` is specified as the URL, but I suspect it may be easier to specify that with a correct Tornado route?

app/access.py
```diff
@@ -72,7 +72,14 @@ def wrapper(self, *args, **kwargs):
                 or "System admin" in self.current_user.permissions
             ):
                 raise tornado.web.HTTPError(401)
-            return method(self, *args, **kwargs)
+            print(method, args, kwargs)
```
stefanv, line 75: Remove

app/access.py
mcoughlin: What does that mean? Right now the app fails with a TypeError if someone messes up.

stefanv: If the route specification in the app factory is specific enough, why does it get sent to the route in the first place?

stefanv: I.e., where does the TypeError come from?

mcoughlin: I don't know what you mean. I just see that error when calling the method. The issue I referenced showed the error message.

mcoughlin:
@stefanv https://github.com/skyportal/skyportal/pull/3604 ok this seems better.

PR #345, opened 2022-10-27
Modify Nginx max request body size (from client) in the config

Theodlz:
This PR allows us to modify the max request body size from the config file of the baselayer.
This is needed in some instances of SkyPortal where we need to upload .fits file (for analysis, or as comment attachments).
Of course, after a first review to figure out if anything should be done differently, it will be followed by an associated PR on SkyPortal main.


mcoughlin (approved):
LGTM

PR #346, opened 2022-12-16
Fix double selenium-requests requirement

bfhealy:
This PR removes a double requirement for `selenium-requests`, keeping the newer version.


guynir42 (approved):

mcoughlin:
@bfhealy Could you quick fix the pointer to gitlab in the workflow? We fixed it in skyportal so can look there.

bfhealy:
@mcoughlin Sure, I'll make that fix.

PR #347, opened 2022-12-21
Increase API request timeout to 30s

bfhealy:
This PR increases the timeout for API requests to 30s.

### DIFF
--- a/services/nginx/nginx.conf.template
+++ b/services/nginx/nginx.conf.template
@@ -71,7 +71,7 @@ http {
   upstream api {
     least_conn;
     {% for p in range(server.dedicated_frontend_processes, [server.processes, server.dedicated_frontend_processes + 1] | max) -%}
-      server 127.0.0.1:{{ ports.app_internal + p }} fail_timeout=20s;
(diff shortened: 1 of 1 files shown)


mcoughlin (commented):
@bfhealy There are a few fail_timeout's in here.

stefanv:
Please read https://nginx.org/en/docs/http/ngx_http_upstream_module.html#fail_timeout.  Specifically, note that making the timeout higher sets *both* detection of timeout *and* the amount of time the server is considered offline before retrying connection.

mcoughlin:
@stefanv We have issues with the permissions checking causing the plotting to timeout (and some reasonable queries also being a bit on the slower side).

stefanv:
All I'm saying is that this may not be the silver bullet you need, since while it will tolerate longer timeouts, it will also *cause* nodes to be offline for longer.

If possible, you can make the slow endpoints async or add more endpoints (might require more CPUs).

mcoughlin:
@stefanv It's true neither the source nor the plot handlers are async currently. Do you want to try that first @bfhealy ?

bfhealy:
@mcoughlin Sure, I'll take a look making that change.

PR #348, opened 2023-01-15
API Call tracking

mcoughlin:
This PR adds API Call tracking.

### DIFF
--- a/app/handlers/base.py
+++ b/app/handlers/base.py
@@ -25,7 +25,14 @@ from ..custom_exceptions import AccessError
 from ..env import load_env
 from ..flow import Flow
 from ..json_util import to_json
-from ..models import DBSession, User, VerifiedSession, bulk_verify, session_context_id
+from ..models import (
+    APICall,
+    DBSession,
+    User,
+    VerifiedSession,
+    bulk_verify,
+    session_context_id,
+)
 
 env, cfg = load_env()
 log = make_log("basehandler")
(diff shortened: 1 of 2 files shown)


profjsb:
Looks good. Question is whether we want to do this logging in the DB itself or another DB. I'm worried about quickly blowing up the DB in size. Can we get an estimate of expected DB growth with the current load if this goes in?

stefanv:
FWIW, all connections are logged in gcloud. This may still be the easiest route to getting the statistics you want, but worth seeing what they store.

mcoughlin:
@profjsb Posting about 2000 API calls, I get a table size of 840 kB. So if we assume an API call a second on production, we would get 36 MB per day.

app/handlers/base.py
```diff
@@ -305,6 +325,25 @@ def error(self, message, data={}, status=400, extra={}):
...
+            session.merge(self.current_user)
+            if hasattr(self.current_user, "created_by_id"):
+                user_id = int(self.current_user.created_by_id)
+            else:
+                user_id = int(self.current_user.id)
+
+            api_call = APICall(
+                user_id=user_id,
+                method=self.request.method,
+                uri=self.request.uri.split("?")[0],
```
guynir42, line 340: I'm not sure, but maybe worth while to save the parameters as well? Maybe limit the number of characters, just in case we get super-long parameter strings, but those parameters could contain lots of useful info. 

Also, consider clipping the beginning of the string (I don't know if URI contains the full address or just the thing after `api/`)

app/handlers/base.py
```diff
@@ -305,6 +325,25 @@ def error(self, message, data={}, status=400, extra={}):
...
+                size=sizeof(data),
+                success=False,
```
guynir42, line 342: I think if you stick something like `self.start_time = time.time()` up in the `prepare` method, you could also get the runtime of the query. That should work unless there's async stuff happening, in which case who knows what will happen... but it is worth trying out and comparing to timing data you get from the side calling the API.

guynir42 (commented):
Looks good. I have a couple of suggestions.

PR #349, opened 2023-01-27
SQLAlchemy 2.0

mcoughlin:
This PR updates SQLAlchemy to 2.0.

### DIFF
--- a/app/handlers/base.py
+++ b/app/handlers/base.py
@@ -142,6 +142,7 @@ class BaseHandler(PSABaseHandler):
             # must merge the user object with the current session
             # ref: https://docs.sqlalchemy.org/en/14/orm/session_basics.html#adding-new-or-existing-items
             session.add(self.current_user)
+            session.bind = DBSession.session_factory.kw["bind"]
             yield session
 
     def verify_permissions(self):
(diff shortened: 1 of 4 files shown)


app/handlers/base.py
```diff
@@ -142,6 +142,7 @@ def Session(self):
             # must merge the user object with the current session
             # ref: https://docs.sqlalchemy.org/en/14/orm/session_basics.html#adding-new-or-existing-items
             session.add(self.current_user)
+            session.bind = DBSession.session_factory.kw["bind"]
```
stefanv, line 145: Looks good, thanks Michael. What is this line for?

mcoughlin: @stefanv I find that the explicit bind is necessary in this new 2.0 setup!

guynir42: what is actually the value of this `kw["bind"]`?

guynir42 (commented):
This is a big step that we should not take lightly. 
Did you try to run this with all of skyportal's tests?

mcoughlin:
@guynir42 It is Engine(postgresql://skyportal:***@localhost:5432/skyportal)

Theodlz:
This looks good. I played around with the associated SkyPortal PR and I didn't notice any misbehavior so far.
At least for the Baselayer part of this SQLAlchemy update, it deserves a LGTM!

mcoughlin:
Ok so @Theodlz is happy and we can most easily test effects of sqlalchemy in skyportal. Will pin this there.

stefanv:
Could you please also patch up the template_app so it runs without warnings under 2.0

PR #354, opened 2023-02-26
social-core DBSession

mcoughlin:
This PR makes the social-core DBSession explicit.

### DIFF
--- a/app/handlers/base.py
+++ b/app/handlers/base.py
@@ -64,7 +64,7 @@ class PSABaseHandler(RequestHandler):
                 ).first()
                 if user is None:
                     return
-                sa = user.social_auth.first()
+                sa = session.query(User).get(user.id).social_auth.first()
                 if sa is None:
                     # No SocialAuth entry; probably machine generated user
                     return user
--- a/app/handlers/base.py
+++ b/app/handlers/base.py
@@ -73,9 +73,10 @@ class PSABaseHandler(RequestHandler):
 
     def login_user(self, user):
         self.set_secure_cookie("user_id", str(user.id))
-        sa = user.social_auth.first()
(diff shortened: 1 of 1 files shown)


mcoughlin:
@Theodlz

Theodlz:
Hi Michael. I would maybe do the same in the method right above this one, where it is also not specified.

Theodlz:
There's 2 things I would change there:
1. Similarly, the user.social_auth to use an explicit session.
2. Maybe right above that, also use an explicit session in the method that gets the user object
3.

Theodlz:
Maybe:

```
def get_current_user(self):
        print("get current user")
        if self.user_id() is None:
            return
        user_id = int(self.user_id())
        oauth_uid = self.get_secure_cookie("user_oauth_uid")
        if user_id and oauth_uid:
            with DBSession() as session:
                user = session.scalars(
                    sqlalchemy.select(User).where(User.id == user_id)
                ).first()
                if user is None:
                    return
                sa = session.query(User).get(user.id).social_auth.first()
                if sa is None:
                    # No SocialAuth entry; probably machine generated user
                    return user
                if sa.uid.encode("utf-8") == oauth_uid:
                    return user
```

mcoughlin:
@Theodlz Ok I think it is the same now except for the extra print.

Theodlz:
Of course. Sorry I forgot to remove the debug print!

mcoughlin:
@Theodlz Anything else? Otherwise I will merge to try it with your onboarding fix to see if this does it.

stefanv (approved):

profjsb:
LGTM. Go ahead and merge after fixing the conflict with base?  I assume there's no easy way to test this?

PR #355, opened 2023-02-27
Social core - catch DBSession errors to proceed with social_core's auth pipeline

Theodlz:
adds on top of what has been added in https://github.com/cesium-ml/baselayer/pull/354 already

### DIFF
--- a/app/handlers/base.py
+++ b/app/handlers/base.py
@@ -59,23 +59,38 @@ class PSABaseHandler(RequestHandler):
         oauth_uid = self.get_secure_cookie("user_oauth_uid")
         if user_id and oauth_uid:
             with DBSession() as session:
+                try:
+                    user = session.scalars(
+                        sqlalchemy.select(User).where(User.id == user_id)
+                    ).first()
+                    if user is None:
+                        return
+                    sa = user.social_auth.first()
+                    if sa is None:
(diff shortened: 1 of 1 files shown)


Theodlz:
@mcoughlin I tested your PR, and spent a solid 2 hours exploring what path the social_core code takes to log in a user (from social_core to its different oauth backend, all the way back to tornado and back to social_core), and here is what I found in just a few words.

The methods in baselayer already exist in social_core BaseHandler. We are overwriting them to check if a user with the user_id and user_oauth_uid from the cookies of whoever is trying to connect exists, in which case we give that existing user to social_core. We avoid a few steps of authenticating the user and running the whole pipeline (that involves interrogating google and using the code in onboarding,...). We had some errors occurring in these 2 methods, blocking the rest of the social_core steps from happening. However, it is fine for these 2 methods to return None instead of a user, which is actually what happens when the person hasn't logged in yet. So having a try catch to simply log the error for us to debug instead of raising it, while allowing the social_core pipeline to proceed with the authentication seems like the expected behavior.

mcoughlin:
@Theodlz sounds good!

mcoughlin (approved):
LGTM. Let me know when you want me to merge and pin @Theodlz

Theodlz:
We can do it now @mcoughlin

PR #357, opened 2023-03-08
Missing SQLAlchemy2 syntax 

Theodlz:
Associated SP PR: https://github.com/skyportal/skyportal/pull/4024


guynir42 (approved):
I don't know how this will make any difference, but it looks fine to me.

PR #359, opened 2023-04-25
Allow for github dependencies

mcoughlin:
This PR allows for dependencies on github branches, a la git+https://github.com/sncosmo/sncosmo.git@master#egg=sncosmo.

### DIFF
--- a/tools/pip_install_requirements.py
+++ b/tools/pip_install_requirements.py
@@ -3,7 +3,6 @@ import sys
 
 import pkg_resources
 from pkg_resources import DistributionNotFound, Requirement, VersionConflict
-from pkg_resources.extern.packaging.requirements import InvalidRequirement
 from status import status
 
 if len(sys.argv) < 2:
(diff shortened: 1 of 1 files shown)


guynir42 (approved):
So, in this case it is up to the user to verify the requirement is installed?

mcoughlin:
@guynir42 Yes.

tools/pip_install_requirements.py
```diff
@@ -40,7 +41,13 @@ def pip(req_files):
 
 try:
     with status("Verifying Python package dependencies"):
-        pkg_resources.working_set.resolve([Requirement.parse(r) for r in requirements])
+        parsed_requirements = []
+        for r in requirements:
+            try:
+                parsed_requirements.append(Requirement.parse(r))
+            except InvalidRequirement:
```
stefanv, line 48: The only thing you can't check is the version, but you can still check whether the package is installed.

mcoughlin: @stefanv Are you thinking of some kind of magic string manipulation to get the package name to pass to Requirement.parse?

stefanv: I believe that `#egg=` is followed by a valid requirement specifier.

mcoughlin: Ok I just split it after that.

PR #360, opened 2023-05-15
Database logging

mcoughlin:
This PR adds a config flag for turning on (or off) database logging.

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -97,6 +97,9 @@ log:
     # log only those that trigger a warning / error
     api_calls: False
 
+    database_echo: False
+    database_echo_pool: False
+
 ports:
     websocket: 64000
     fake_oauth: 63000
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -97,8 +97,11 @@ log:
     # log only those that trigger a warning / error
     api_calls: False
 
-    database_echo: False
-    database_echo_pool: False
+    # if true, log database interactions
+    database: False
+
+    # set the debugging level (debug or info)
+    database_pool: False
 
 ports:
     websocket: 64000
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -100,10 +100,10 @@ log:
     # SQL logging
     # see e.g. https://docs.sqlalchemy.org/en/20/core/engines.html#configuring-logging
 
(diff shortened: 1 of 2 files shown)


config.yaml.defaults
```diff
@@ -97,6 +97,9 @@ log:
     # log only those that trigger a warning / error
     api_calls: False
 
+    database_echo: False
```
stefanv, line 100: Please add comments explaining what these do.

"echo" is not very descriptive; perhaps use `log.database` and `log.database_pool`?

stefanv:
Sorry, I don't mean to be annoying, but those comments cannot help a user understand what is being logged.

stefanv:
From the page you linked:

```
sqlalchemy.engine - controls SQL echoing.
sqlalchemy.pool - controls connection pool logging.
```

So, while I haven't verified locally, looks like the first echoes SQL commands, and the second logs connections made to the database?

config.yaml.defaults
```diff
@@ -97,10 +97,11 @@ log:
     # log only those that trigger a warning / error
     api_calls: False
 
-    # if true, log postgres database queries
+    # if true, enable SQL echoing.
     database: False
 
-    # set the debugging level (debug or info)
+    # if true, enable connection pool logging
+    # can also set the debugging level (debug or info)
```
stefanv, line 104: So can this flag be anything in `True, False, "debug", "info"`?

mcoughlin: I checked and True, False, and "debug" work. "info" did not for me so I took that off.

config.yaml.defaults
```diff
@@ -97,6 +97,15 @@ log:
     # log only those that trigger a warning / error
     api_calls: False
 
+    # SQL logging
+    # see e.g. https://docs.sqlalchemy.org/en/20/core/engines.html#configuring-logging
+
+    # if True or "debug", enable SQL echoing.
+    database: False
+
+    # if true or "debug", enable connection pool logging
```
stefanv, line 106: `True`? And if `True` is just a synonym for `debug`, I'd remove that latter option. Preferably only one way to spell something! Then I'll merge.

mcoughlin: That does seem better.

stefanv:
Thanks @mcoughlin!

PR #361, opened 2023-05-24
Support JSON serialization of ObjectId

Theodlz:
This PR adds the support for ObjectId serialization. For any application using baselayer and interacting with MongoDB, not supporting ObjectId serialization means having to add acustom method pre serialization to convert it to string. It would be helpful to support it in Baselayer directly.

### DIFF
--- a/app/json_util.py
+++ b/app/json_util.py
@@ -2,6 +2,7 @@ from datetime import date, datetime
 
 import simplejson as json
 from arrow.arrow import Arrow
+from bson import ObjectId
 from sqlalchemy_utils import PhoneNumber
 
 data_types = {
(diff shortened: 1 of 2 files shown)


Theodlz:
@mcoughlin

mcoughlin (approved):
Thanks @Theodlz. This will fix the issue we saw in fritz this morning.

mcoughlin (approved):
Thanks @Theodlz. This will fix the issue we saw in fritz this morning.

mcoughlin:
@Theodlz can you fix the linting please?

Theodlz:
> @Theodlz can you fix the linting please?

Of course, doing it right now.

stefanv:
Having pymongo as a dependency just to handle serialization is a bit much. Presumably, we can just look at `__class__` or similar, and if it matches call the appropriate method.

Theodlz:
> Having pymongo as a dependency just to handle serialization is a bit much. Presumably, we can just look at `__class__` or similar, and if it matches call the appropriate method.

Sure, let me do that right now.

Theodlz:
@stefanv is that the approach you had in mind?

stefanv:
Yup, perfect.

PR #363, opened 2023-05-30
Set ssl certificates from config

mcoughlin:
This PR makes it possible to set set ssl certificates from the config (to ease updating the code when you modify this file).

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -26,6 +26,11 @@ server:
     # this should be set to False.
     ssl: False
 
+    # If you are using certificates, you can specify them here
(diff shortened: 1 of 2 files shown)


config.yaml.defaults
```diff
@@ -26,6 +26,11 @@ server:
     # this should be set to False.
     ssl: False
 
+    # If you are using certificates, you can specify them here
+    ssl_use_certificate: False
```
stefanv, line 30: Please add a link to the nginx docs, showing what these arguments should be. Then good to go.

mcoughlin: Added!

PR #365, opened 2023-07-16
Add kwargs to to_json

mcoughlin:
This PR adds a kwargs to to_json, useful for using to_json as a drop-in replacement for json.dumps (which it is, but will benefit from taking the various optional args).

### DIFF
--- a/app/json_util.py
+++ b/app/json_util.py
@@ -52,5 +52,5 @@ class Encoder(json.JSONEncoder):
         return json.JSONEncoder.default(self, o)
 
 
-def to_json(obj):
-    return json.dumps(obj, cls=Encoder, indent=2, ignore_nan=True)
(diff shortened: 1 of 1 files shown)


mcoughlin:
@Theodlz useful for your jinga2 templating.

app/json_util.py
```diff
@@ -52,5 +52,5 @@ def default(self, o):
         return json.JSONEncoder.default(self, o)
 
 
-def to_json(obj):
-    return json.dumps(obj, cls=Encoder, indent=2, ignore_nan=True)
+def to_json(obj, **kwargs):
```
stefanv, line 55: Pop indent from kwargs with 2 as the default, if it is not found. The rest looks good.

mcoughlin: Yes added

PR #366, opened 2023-12-08
Refactor Sessions to clarify their usage

stefanv:
DBSession has been renamed to HandlerSession.

The following exist now:
- HandlerSession: a DB session scoped to the current handler; should therefore only be used inside handlers.
- ThreadSession: a DB session scoped to the current thread; should be used outside of handlers, e.g. when firing off tasks in threads, or inside of services.

A special HandlerSession, with knowledge of the current user, and that verifies permissions on commit, is available as `self.Session` on the base handler. This is the *preferred way of accessing* HandlerSession.

To make it easier to get hold of the current engine, the Sessions each have a `.engine` attribute: `HandlerSession.engine`.

This avoids having to access the engine as
`HandlerSession.session_factory.kw["bind"]`.


stefanv:
@Theodlz Just to confirm that we decided not to take this route, so there are no ideas from this PR that need to be saved?

PR #369, opened 2023-12-09
WIP - Service plugins

Theodlz:
Just a quick experiment with plug-and-play microservices. Other instances of SkyPortal (such as Icare, used by GRANDMA) have some custom microservices. It's a little awkward that they have to maintain their special architecture to get just one microservice running.

This is just an idea of how we could allow one to add as many custom microservices as needed with just a few lines added to baselayer.

And, here's the demo repo that shows what a git repo for a custom microservice needs to look like.

Definitely a WIP, but it works perfectly as is.

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -156,3 +156,5 @@ security:
 #   - interval: 5
 #     script: tools/5_minute.py
 #     limit: ["01:00", "02:00"]
+
+plugins:
--- a/tools/setup_services.py
+++ b/tools/setup_services.py
@@ -1,12 +1,59 @@
 import os
 from collections import Counter
 from os.path import join as pjoin
+import subprocess
 
 from baselayer.app.env import load_env
 from baselayer.log import make_log
 
 log = make_log("baselayer")
 
+def download_plugin_services():
+    env, cfg = load_env()
+
+    services_path = cfg["services.paths"]
+
+    plugins = cfg.get("plugins", {})
+    plugin_services = []
+
+    log(f"Discovered {len(plugins)} plugins")
+
+    for plugin_name, plugin_info in plugins.items():
+        if "url" not in plugin_info:
+            log(f"Skipping plugin {plugin_name} because it has no URL")
+            continue
+
+        git_repo = plugin_info["url"].split(".git")[0].split("/")[-1]
+        plugin_path = pjoin(services_path[-1], git_repo)
+
+        if os.path.exists(plugin_path):
+            if os.path.exists(pjoin(plugin_path, ".git")):
+                log(f"Updating plugin {plugin_name}")
+                _, stderr = subprocess.Popen(
+                    f"cd {plugin_path} && git pull",
+                    shell=True,
+                    stdout=subprocess.PIPE,
+                    stderr=subprocess.PIPE,
+                ).communicate()
+                if stderr and len(stderr) > 0:
+                    log(f"Error updating plugin {plugin_name}: {stderr}")
+                    continue
+            else:
+                log(f"Skipping plugin {plugin_name} because a microservice with the same name exists at {plugin_path}")
+            continue
+        else:
+            log(f"Cloning plugin {plugin_name}")
+            _, stderr = subprocess.Popen(
+                f"cd {services_path[-1]} && git clone {plugin_info['url']}",
+                shell=True,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+            ).communicate()
+            if stderr and len(stderr) > 0:
+                log(f"Error cloning plugin {plugin_name}: {stderr}")
+                continue
+        plugin_services.append(plugin_name)
+    return plugin_services
 
 def copy_supervisor_configs():
     env, cfg = load_env()
@@ -56,4 +103,6 @@ def copy_supervisor_configs():
 
 
 if __name__ == "__main__":
+    download_plugin_services()
+    print()
     copy_supervisor_configs()


stefanv:
What does a plugin do vs a microservice?

Theodlz:
Hi @stefanv,

It's the same thing! The idea here is just to allow folks to add custom micro services to their app without having to actually merge the service to the app's repo. 

A good example is using Fink broker in SkyPortal to get new. Doesn't make sense to add it to SkyPortal but one might one to add it to its own instance.

This would allow that. Happy to change the name/terminology used to call this new feature.

The reason why I called it plug-ins is that this could be extended to not just micro services but maybe front end and API one day in the future.

stefanv:
Ah, in that case I wonder if we shouldn't just add the submodule functionality to the services?

I'd also like to move port definitions and other configuration so that each service definition happens in one place.

stefanv:
If you use the submodule path, the local path version also works just fine; you just need to do a checkout --recursive on your repo.

Theodlz:
> If you use the submodule path, the local path version also works just fine; you just need to do a checkout --recursive on your repo.

Sure, though I think it's always good to do some validation, i.e warning, failing, or auto initializing submodules  if baselayer starts and it's not done yet.

tools/setup_services.py
```diff
@@ -1,15 +1,396 @@
...
+    try:
+        v = version.parse(version_string)
+        spec = SpecifierSet(requirement_string)
+        return v in spec
+    except Exception as e:
+        log(f"Error validating version: {e}")
+        return False
+
+
+def validate_plugin_compatibility(plugin_name: str, plugin_path: str):
```
stefanv, line 81:
```suggestion
def validate_plugin_compatibility(plugin_name: str, plugin_path: str) -> bool:
```

tools/setup_services.py
```diff
@@ -1,15 +1,396 @@
...
+    plugin_config = read_plugin_config(plugin_path)
+    name, version_requirement = get_plugin_compatible_version(plugin_config)
+    if not (name and version_requirement):
+        return True
+    try:
+        mod = __import__(name)
```
stefanv, line 87:
```suggestion
        mod = importlib.import_module(name)
```

tools/setup_services.py
```diff
@@ -1,15 +1,396 @@
...
+        installed_version = mod.__version__
+
+        if not validate_version(installed_version, version_requirement):
+            log(
+                f"Plugin {plugin_name} is incompatible: required {version_requirement}, found {installed_version}. Skipping."
```
stefanv, line 92: "incompatible" is a bit ambiguous here; what is incompatible with what.

Would be helpful if we could say something like:

"Plugin myplugin requires skyportal x, but this is skyportal y. Skipping."

tools/setup_services.py
```diff
@@ -1,15 +1,396 @@
...
+
+
+def read_plugin_config(plugin_path):
+    # read the config (pyproject.toml) for the plugin
+    config_path = pjoin(plugin_path, "pyproject.toml")
+    with open(config_path, "rb") as f:
+        return tomli.load(f)
+
+
+def get_plugin_compatible_version(plugin_config: dict):
```
stefanv, line 39: Annotate return type

tools/setup_services.py
```diff
@@ -1,15 +1,396 @@
...
+    # if the plugin specifies a "tool.<name>" section with
+    # a version requirement, retrive it
```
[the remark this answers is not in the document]
stefanv: This doesn't seem like a complete description; we return two things here, what are they?

tools/setup_services.py
```diff
@@ -1,15 +1,396 @@
...
+        installed_version = mod.__version__
+
+        if not validate_version(installed_version, version_requirement):
+            log(
+                f"Plugin {plugin_name} is incompatible: required {version_requirement}, found {installed_version}. Skipping."
+            )
+            return False
+    except ImportError:
+        log(
+            f"Could not find package {name} which plugin {plugin_name} depends on. Skipping."
```
stefanv, line 97:
```suggestion
            f"Plugin {plugin_name} depends on missing package {name}. Skipping."
```

stefanv: I'd have to check more carefull, but is this a general check or only for skyportal? If more general, do we want to provide the ability to install packages that the plugin depends on, if they are not already present on the system?

tools/setup_services.py
```diff
@@ -1,15 +1,396 @@
...
+    if plugin_config["project"]["name"] != plugin_name:
+        log(
+            f"Plugin {plugin_name} has a different name in its config ({plugin_config['project']['name']}). Skipping."
+        )
+        return False
+
+    return True
+
+
+def run_git_command(args, plugin_path, plugin_name):
```
stefanv, line 110: Other function signatures are decorated; want to decorate throughout?

stefanv, line 110: Other function signatures are decorated; want to decorate throughout?

tools/setup_services.py
```diff
@@ -1,15 +1,396 @@
...
+    """
+    Run a git command in the specified plugin path.
+
+    Args:
```
stefanv, line 114: Do we use Google docstring format in the rest of skyportal? If not, make this consistent.

tools/setup_services.py
```diff
@@ -1,15 +1,396 @@
...
+        args (list): Git command, e.g. ['checkout', 'main']
+        plugin_path (str): Directory where the command runs
+        plugin_name (str): Used in error logs
+
+    Returns:
+        stdout_lines (list of str): Output lines from stdout
```
stefanv, line 120: Why does stdout get split into lines, but stderr is provided as a str?

tools/setup_services.py
```diff
@@ -1,15 +1,396 @@
...
+    os.makedirs(plugins_path, exist_ok=True)
+
+    plugins = cfg.get("plugins", {})
+    plugin_services = []
+
+    log(f"Discovered {len(plugins)} plugins")
+
+    for plugin_name, plugin_info in plugins.items():
+        if "url" not in plugin_info:
+            log(f"Skipping plugin {plugin_name} because it has no URL")
```
stefanv, line 158:
```suggestion
            log(f"Skipping plugin {plugin_name}: no URL found in config")
```

Maybe provide some guidance here even, as to where and how user should specify config.

tools/setup_services.py
```diff
@@ -1,15 +1,396 @@
...
+        return update_to_branch(plugin_name, plugin_path, branch)
+
+
+def clone_new_plugin(plugin_name, plugin_info, plugin_path):
+    """Clone a new plugin repository."""
+    branch = plugin_info.get("branch", "main")
+    version_tag = plugin_info.get("version")
+    sha = plugin_info.get("sha")
+
+    log(f"Cloning plugin {plugin_name}")
```
stefanv, line 209: Instead of doing a deep clone, can we do a shallow clone of exactly the revision specified? If so, should be able to get rid of the three `update_to_` functions.

tools/setup_services.py
```diff
@@ -19,6 +400,15 @@ def copy_supervisor_configs():
             ]
             services.update({s: pjoin(path, s) for s in path_services})
 
+    all_plugins_path = cfg.get("services.plugins_path", "./plugins")
+    for p in activated_plugins:
+        services.update({p: pjoin(all_plugins_path, p)})
+
+    # TODO (in a future PR): loop over all services, check if they are a git submodule or not
+    # if they are a submodule make sure they are initialized and updated
+    # this should be discussed, it does not seem necessary as soon as we have the
```
stefanv, line 409: I think user can just take care of submodule updates, then plugin can be specified as file

config.yaml.defaults
```diff
@@ -86,6 +86,7 @@ services:
     paths:
       - ./baselayer/services
       - ./services
+    plugins_path: ./plugins
```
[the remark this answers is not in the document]
Theodlz: Yes, I think it's good to differentiate where plugins that we clone vs other microservices go. This makes it much cleaner and obvious for the user what is a plugin, and what is a built-in microservice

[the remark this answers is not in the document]
Theodlz: I did this as I found myself very quickly annoyed at having to scroll through all of skyportal's microservices to find mind. Less ambiguity the better imo

tools/setup_services.py
```diff
@@ -1,15 +1,396 @@
...
+redirect_stderr=true
+"""
+    # Write the configuration to a file
+    with open(f"{service_path}/supervisor.conf", "w") as f:
+        f.write(supervisor_conf_template)
+
+
+def read_plugin_config(plugin_path):
+    # read the config (pyproject.toml) for the plugin
+    config_path = pjoin(plugin_path, "pyproject.toml")
```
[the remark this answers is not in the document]
Theodlz: Yes, so the idea is to conform to the python project layout standards. Now folks use a pyproject.toml for a python project. This way in theory, a plugin can work on its own too and such.

requirements.txt
```diff
@@ -20,3 +20,4 @@ phonenumbers>=8.12.15
 python-slugify>=4.0.1
 numpy>=1.21.4
 webdriver-manager>=4.0.1
+tomli
```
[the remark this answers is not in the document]
stefanv: Right, so include that dependency only for less than 3.11, and then do a conditional import of one or the other.

tools/setup_services.py
Theodlz: I think @thomasculino needed to parse multiple lines for some commands, but we should standardize that. We'll talk about it

tools/setup_services.py
Theodlz: The advantages I see w/ using the plugin system for submodules is version validation + in general giving some standard place in the config for params that these submodules should read from. But we don't need to do any branch/sha/version management. Any thoughts?

tools/setup_services.py
Theodlz: will make these changes along with the doc strings

tools/setup_services.py
Theodlz: Agreed, we'll clarify all the return messages.

tools/setup_services.py
Theodlz: it's general, the plugins's pyproject.toml tool.X is where we get the name of the software that the plugin can be installed for (.e.g tool.skyportal).

I don't see why skyportal vs generic makes a different about whether or not we want to install extra packages but let me answer that question anyway: yes I think we want to allow that. But that requires some careful thinking and should somehow be integrated in our current python dependencies check to see if things are compatible. So I would ignore it in this first PR and have one dedicated to enabling the installation of additional dependencies.

requirements.txt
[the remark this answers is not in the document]
Theodlz: Will double check, but I doubt it

tools/setup_services.py
stefanv: I guess I'm thinking: the default services are essentially exactly the same as plugins, except they don't do version checking. So, a user can add submodules, check them out with their project, and augment the services path. We can update services to check skyportal version, if a `pyproject.toml` is available. That way we have one unified pipeline for both, and `plugins` just become a mechanism for checking out plugins from remote locations.

requirements.txt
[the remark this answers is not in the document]
Theodlz: ok done

tools/setup_services.py
```diff
@@ -1,15 +1,396 @@
...
+
+def get_plugin_compatible_version(plugin_config: dict):
+    # if the plugin specifies a "tool.<name>" section with
+    # a version requirement, retrive it
+    if "tool" not in plugin_config:
+        log("Plugin config does not contain 'tool' section")
+        return None, None
+    # there should only be one key in tool, which is the name of the package
+    # for which we are installing the plugin
+    if len(plugin_config["tool"]) != 1:
```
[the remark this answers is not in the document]
Theodlz: done

tools/setup_services.py
Theodlz: done

tools/setup_services.py
Theodlz: done

tools/setup_services.py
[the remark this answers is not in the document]
Theodlz: done

tools/setup_services.py
Theodlz: done

tools/setup_services.py
Theodlz: done

tools/setup_services.py
Theodlz: done

tools/setup_services.py
Theodlz: still a TODO

Theodlz: well examples are given in the config default, so maybe that's enough?

stefanv:
Great, thanks for reworking this! I will review ASAP, but remember also to update the docs please.

Theodlz:
We'll ping you here and on slack when it's ready for another review

Theodlz:
@stefanv looks like the precommit CI is making some really weird linting choices ngl

doc/extending.md
```diff
@@ -61,3 +61,28 @@ can run the templating manually:
 ```
 ./baselayer/tools/fill_conf_values.py --config="config.yaml" static/js/component.jsx.template
 ```
+
+## Adding extrernal services
+
+If you want to add external services to your application,
```
Theodlz, line 67: I think we should explain what *is* an external service before saying how to add one. Explaining the concept, etc.

doc/extending.md
```diff
@@ -61,3 +61,28 @@ can run the templating manually:
...
+you can do so by adding them to the `config.yaml` file under the
+`services.external` key. Each service should have a unique name and a set of parameters that define how to connect to it. For example:
```
Theodlz, line 69: I don't think there are any parameters about how to "connect" to it, the wording could be a bit confusing given that there are actually networking concerns that are separate from that. I would rephrase it to make it clear that the config here let's baselayer pull the external service from a separate github repo. Also would be good to mention the versioning (branch+sha or version), which is omitted at the moment

doc/extending.md
```diff
@@ -61,3 +61,28 @@ can run the templating manually:
...
+    another_service:
+        url: "https://github.com/another_service.git"
+        version: v0.1.0
+        params: {}
+```
+
+Your script will then run when the application starts.
+
+This allows you to easily integrate external services into your
+application without hardcoding configuration values in your codebase.
```
Theodlz, line 88: I don't understand that sentence, I don't think the big advantage is to avoid hardcoding configs, it's that you don't need to fork and diverge from the main repo rather, so this is misleading. With the external services we can extend on the application without changing the application itself

tools/setup_services.py
```diff
@@ -441,15 +458,17 @@ def update_or_clone_plugin(
...
         return update_or_clone_plugin_by_tag(url, version_tag, plugin_name, plugin_path)
     elif branch and sha:
-        log(f"Updating {plugin_name} to branch {branch} at SHA {sha}")
         return update_or_clone_plugin_by_branch(
             url, branch, sha, plugin_name, plugin_path
         )
-    else:
+    elif os.path.isdir(plugin_path):
         return True
+    # TODO: if it exists but it's an uninitialized git submodule
```
Theodlz, line 468: maybe let's remove this TODO, feels like submodules could be handled in a separate PR. In general I think we shouldn't try to do EVERYTHING imaginable in this one, and better merge a first version we are happy with.

Theodlz:
@thomasculino  could you maybe create a new repo that is a copy/paste of my boom-specific plugin, wipe the main method clean and just have some periodic print in there? Something simple we can mention in the docs here as an example of an external service, and so stefan can try it out.

doc/usage.md
```diff
@@ -55,6 +56,26 @@ services:
     enabled:
       - cron
     disabled: '*'
+    external:
+```
+
+External microservices can also be added to the `config.yaml` file
```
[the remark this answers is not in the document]
thomasculino: Should we keep only one file where we document these external services? It seems a bit awkward to do it in two locations if the content is pretty much the same.

doc/extending.md
```diff
@@ -61,3 +61,30 @@ can run the templating manually:
...
+  external:
+    my_service:
+        url: "https://github.com/my_service.git"
+        branch: main
+        sha: specific_commit_sha
+        params: {}
+    another_service:
+        url: "https://github.com/another_service.git"
+        version: v0.1.0
+        params: {}
```
Theodlz, line 85: As an example of how to add `params` (and to show it's optional), maybe it would make sense to have the first microservice not specify it, and the second one specify it but not empty (with some actual keys)?

doc/extending.md
```diff
@@ -61,3 +61,30 @@ can run the templating manually:
...
+```
+
+You must provide the `url` of the GitHub repository. To target a specific version of the service, you can optionally include both the `branch` and `sha`, or use the `version` field to refer to a particular release. Additional configuration options can be passed through the `params` dictionary, which the service may use during setup.
```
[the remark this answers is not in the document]
Theodlz: `you can optionally include` doesn't seem correct. You HAVE to provide a branch+sha OR a version to add an external service, right?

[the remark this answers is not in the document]
Theodlz: `Additional configuration options can be passed through the `params` dictionary, which the service may use during setup.` I mean these are just keys the service may use, idk if it's necessarily at `setup` time.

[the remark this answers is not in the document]
Theodlz: Maybe here is where it's good to mention:
- external services should provide a `main.py`, or if there entry point is different they have to provide a `supervisord.conf` (which we otherwise auto generate to point to main.py
- that they can provide a pyproject.toml file that contains the service's name (which is the same name that should be used in the config), and a `tool.X` section where `X` is the name of the application that is built on top of baselayer. And, that this section should include a `version` if specified, which is a version requirement, to make sure they only are registered and run with versions of `X` that they support.
- that they can provide a `config.yaml.defaults` that contain a default configuration for the service, useful for external services that have default parameters (in `params:`), which also acts as an example of how to add the external service to an application.

doc/extending.md
```diff
@@ -61,3 +61,30 @@ can run the templating manually:
...
+
+The external service will then be initialized and registered in supervisor, alongside other services, provided it is correctly configured and compatible with the application.
```
Theodlz, line 90: See for example here you talk about "compatibility" but it's never defined before, hence my previous comment

doc/usage.md
```diff
@@ -55,13 +56,35 @@ services:
...
 ```
 
 Sometimes, the supervisor configuration needs information from the
 configuration file, therefore `supervisor.conf` can instead be
 provided as `supervisor.conf.template`, which will be compiled before
 launching. See, e.g., `services/dask`.
 
+External services can also be used to pull a custom microservice from a Github repository.
+
+They can also be added to the `config.yaml` file under the `services.external` key. Each service should have a unique name and a set of parameters that define what specific repository you want to pull. For example:
```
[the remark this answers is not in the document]
Theodlz: The repeated `can also` did not make much sense in this context, especially how it was repeated. That is my main comment, other changes aren't mandatory.

doc/usage.md
```diff
@@ -55,13 +56,35 @@ services:
...
+
+```
+services:
+  paths:
+    - ./baselayer/services
+    - ./services
+  enabled:
+    - cron
+  disabled: '*'
```
Theodlz, line 78: Not really needed to show these here.

doc/usage.md
Theodlz, line 1: Since the way these work is detailed in the other file in greater detail, wondering if this one could just mention the external services' existence in a short paragraph and just refer to the other page. I'd say for now let's address my comments to clarify things and let's see what Stefan thinks. I feel like a bit of repetition isn't necessarily a bad thing actually, as long as it does not create more confusion.

Theodlz (commented):
Just some comments about the docs Thomas, nice job overall but a number of things that are part of the new plugin system (auto generated supervisor.conf, structure required for the plugins with pyproject.toml, optional config.yaml.defaults, ...) are omitted. These need to be in there.

Also not reflected in my individual suggestions and comments, but let's keep in mind that things like the optional config.yaml.defaults, automated supervisor.conf, and optional pyproject.toml are not external services specific at all!!! We now enforce that on all services, so I think it should move out of the external service-specific sections and be mentioned as micro-service registration feature(s) in general, and the external services stuff only explain what is specific to that system.

doc/usage.md
thomasculino: These were already there, I just added the disable key

Theodlz:
Hey Stefan, I think this is ready for another round of review. You can test adding plugins by adding https://github.com/Theodlz/boom-skyportal-plugin, or Thomas' https://github.com/thomasculino/baselayer_external_service_template (the latter should be moved to the cesium organization imo).

config.yaml.defaults
```diff
@@ -92,10 +92,21 @@ services:
...
 
     # A list of service names, or '*'
+    # You can use disabled to disable specific services, or '*' to disable them all
     disabled:
       - dask
 
-    # You can use disabled to disable specific services, or '*' to disable them all
+    # Here you can add services that are not part of baselayer, but
+    # come from another github repository. You need to specify URL,
+    # branch+sha or version (tag), and optional . For example:
```
stefanv, line 101: "and optional..." <- incomplete

config.yaml.defaults
```diff
@@ -92,10 +92,21 @@ services:
...
+    # - my_service:
+    #     url: <url/to/git/repo>
+    #     branch: <branch_name>
+    #     sha: <sha_of_commit>
```
stefanv, line 105: `branch` and `sha` could probably be same specified. Like, e.g., for pre-commit you would specify an action to run with `https://github.com/some/action@branch` or `https://github.com/some/action@abc123`.

config.yaml.defaults
```diff
@@ -92,10 +92,21 @@ services:
...
+    # - my_service:
+    #     url: <url/to/git/repo>
+    #     version: <tagged_version>
```
stefanv, line 108: This too. Generic git term: `ref` (`reference`)

config.yaml.defaults
```diff
@@ -156,3 +167,5 @@ security:
 #   - interval: 5
 #     script: tools/5_minute.py
 #     limit: ["01:00", "02:00"]
+
+plugins: {}
```
stefanv, line 171: Why plugins, with external services already defined?

doc/extending.md
stefanv: :+1: I can't see higher up in that file, but we should check that the concept of a *service* is well defined. Also make very clear that external services are identical to services, they just live in remote repos.

doc/extending.md
```diff
@@ -61,3 +61,84 @@ can run the templating manually:
...
 ```
+
+## Adding external services
+
+External services are microservices that you can pull from GitHub and run as part of your application. This is useful for integrating third-party services or custom scripts.
+
+If you want to add external services to your application,
+you can do so by adding them to the `config.yaml` file under the `services.external` key. This allows baselayer to pull the external service from a GitHub repository.
+
+The configuration in the `config.yaml` file should look like this:
```
stefanv, line 72:
```suggestion
The configuration in the `config.yaml` file would look like this:
```

doc/extending.md
```diff
@@ -61,3 +61,84 @@ can run the templating manually:
...
+
+You must provide the `url` of the GitHub repository, as well as a version of the service. To specify the version, you are required to either:
+
+- Provide both the `branch` and `sha`,
+  **or**
+- Use the `version` field to reference a specific release.
+
+Additional configuration options can be passed through the `params` dictionary, which the service uses.
+
+The external service will then be initialized and registered in `supervisor` alongside other services, provided it is correctly configured and compatible with the application.
```
stefanv, line 96:
```suggestion
The external service will then be started and managed by `supervisord` alongside other services.
```

doc/extending.md
```diff
@@ -61,3 +61,84 @@ can run the templating manually:
...
+Compatibility can be enforced via version constraints declared in the service’s optional pyproject.toml. See [External Service Requirements](#external-service-requirements) for details.
```
stefanv, line 97:
```suggestion
```

doc/extending.md
```diff
@@ -61,3 +61,84 @@ can run the templating manually:
...
+
+### External Service Requirements
+
+To work correctly with the application, external services should follow these conventions:
+
+#### Entry Point
+
+- The service should include a `main.py` file as its entry point. In that case, providing a `supervisor.conf` can be omitted and will be auto-generated, referring to `main.py`.
```
stefanv, line 105:
```suggestion
- The service should include a `main.py` file as its entry point.
```

stefanv: I think just force this as a default requirement, with a note that you *can* override it by specifying `supervisord.conf` if you really want to.

doc/extending.md
```diff
@@ -61,3 +61,84 @@ can run the templating manually:
...
+- If the entry point differs, a `supervisord.conf` must be provided in the repository, pointing to the correct entry point.
+
+#### Project Metadata and Compatibility
+
+External services may include a `pyproject.toml` file to provide metadata and compatibility information. This is especially important when multiple applications (or versions) are built on top of Baselayer.
```
stefanv, line 110: What does this mean: "especially important when multiple applications (or versions) are built on top of Baselayer"?

doc/extending.md
```diff
@@ -61,3 +61,84 @@ can run the templating manually:
...
+
+In this file:
+
+- The name of the service must be specified, matching the external service name in `config.yaml`.
+
+- It should define a [tool.`<application-name>`] section, where `<application-name>` is the name of the application you are adding the external microservice to. This section may include a `version` field to declare which versions of the application the service is compatible with. If the version requirement is not met, the external service will not be registered.
+
+For example, an external service's `pyproject.toml` may look like:
+
+```toml
```
stefanv, line 120: Show a template example first, then discuss all possible keys. You can choose whether or not to include those keys in the template.

doc/extending.md
```diff
@@ -61,3 +61,84 @@ can run the templating manually:
...
+[project]
+name = "my_service"
+version = "0.1.0"
+description = "A micro-service to add to an application built on top of baselayer"
+authors = [
+  { name = "John Doe", email = "john.doe@example.com" }
+]
+
+[tool.myapp]
```
stefanv, line 129: Probably better to specify this as something like:

```
compatible-with = [...]
```

or similar, instead of requiring a key named after a package.

doc/extending.md
```diff
@@ -61,3 +61,84 @@ can run the templating manually:
...
+version = ">=1.2.0, <2.0.0"
+```
+
+#### Default Configuration
+
+A `config.yaml.defaults` file can be provided, which is helpful if the service needs to be configured and expects parameters to be specified in `config.yaml`. It also serves as an example of how to integrate the service into an application:
```
stefanv, line 135:
```suggestion
A `config.yaml.defaults` file can be provided to set service configuration defaults.
These values can be overridden in the app's `config.yaml` under the `services.external.<name-of-this-service>.config` keys:
```

Why params and not config?

stefanv (commented):
Reviewed the docs first, to ensure we are in agreement on higher level stuff, then will do a pass over the code :pray:

config.yaml.defaults
thomasculino: Deleted, it was a mistake.

Theodlz:
@stefanv I believe this is ready for another round, Thomas made a lot of changes and answered your comments

doc/extending.md
```diff
@@ -61,3 +61,84 @@ can run the templating manually:
...
+#### Default Configuration
+
+A `config.yaml.defaults` file can be provided to set service configuration defaults.
+These values can be overridden in the app's `config.yaml` under the `services.external.<name-of-this-service>.config` keys:
+
+```yaml
+services:
+  external:
+    my_service:
+      version: v0.1.0
```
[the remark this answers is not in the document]
Theodlz: No but that's the version to pull the plugin. You can specify a version (that's a tagged version on git), OR a sha.

app/config.py
```diff
@@ -112,6 +113,33 @@ def load_config(config_files=[]):
     all_configs = [cfg for cfg in all_configs if os.path.exists(os.path.normpath(cfg))]
     all_configs = [os.path.abspath(Path(c).absolute()) for c in all_configs]
 
-    cfg = Config(all_configs)
+    cfg = Config(all_configs, verbose=False)
```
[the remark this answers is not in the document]
Theodlz: I think the issue here is that we needed to load the configs first to figure out WHERE the services are located, so we can load them and their services. Forcing us to load the non-services configs first to discover the services.

PR #371, opened 2023-12-15
2023-12 refresh

stefanv:


Theodlz:
Just added some other changes to update GAs, minus the selenium pinning which is required for the tests to pass until we make changes to baselayer itself to work with the next version of it.

PR #372, opened 2023-12-15
Refresh 2023-12 - debug

Theodlz:
Just a PR to get the GAs to run and try to fix them. Will commit anything back to @stefanv's PR if successful

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -14,7 +14,7 @@ jobs:
     if: github.repository_owner == 'cesium-ml'
 
     steps:
-      - uses: actions/cache@v2
+      - uses: actions/cache@v3
         with:
           path: |
             ~/.cache/pip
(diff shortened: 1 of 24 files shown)


Theodlz:
@stefanv should we also consider running the actions on a version of python that is more recent, say 3.10 or above? We know at least that 3.10 works perfectly for SkyPortal for instance

Theodlz:
maybe also pin a version of selenium? looks like here the GA we just install the latest. SP pins `selenium==4.8.3`

Theodlz:
Oh and also, the node setup script we used is apparently deprecated (that broke the tests on SP and so many repos a little over a month ago). Let's maybe use the GA that installs node instead.

Theodlz:
@stefanv pinning selenium to 4.8.3 did the trick. Ideally we might want to make sure it works with a more recent version, so I can look into their changelog maybe

stefanv:
Fantastic, thanks, that makes the upgrade path a lot clearer; I think I already know what to change with the new selenium.

Theodlz:
@stefanv Great! Do you want me to remove the selenium-related changes here, and move that PR to point to your branch so we can merge the rest of the changes to your PR?

stefanv:
If you can just push the relevant commit to my PR, that should do the trick, thank you!

Theodlz:
> If you can just push the relevant commit to my PR, that should do the trick, thank you!

Done! Sorry for the delay

PR #373, opened 2024-01-11
webdriver-manager

mcoughlin:
This PR adds webdriver-manager to deal with GeckoDriver (see https://stackoverflow.com/questions/46753393/how-to-run-headless-firefox-with-selenium-in-python, as my upgrade to Sonoma broke it).

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -65,6 +65,12 @@ jobs:
           cd baselayer
 
           sudo apt update -y
+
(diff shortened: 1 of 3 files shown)


stefanv:
Adding the dependency is not a big deal, but why not just specify the path to Firefox directly, avoiding it?

mcoughlin:
@stefanv This seemed like an easy cross-platform thing.

Issue #374, opened 2024-01-18
Issue installing dependencies


Theodlz:
The last PR merged to main: https://github.com/cesium-ml/baselayer/pull/324 creates some issues upstream.

It imports packaging in the python script that we use to actually install the dependencies. So that forces you to do a pip install first on your own, which somewhat defeats the purpose of the `dependencies` block of code calls by all of the `make` commands.

I'm just wondering if we shouldn't just remove it from the requirements.txt, and instead change:
```make
dependencies: README.md
	@baselayer/tools/check_app_environment.py
	@PYTHONPATH=. python baselayer/tools/pip_install_requirements.py baselayer/requirements.txt requirements.txt
	@./baselayer/tools/silent_monitor.py baselayer/tools/check_js_deps.sh
```

to:
```make
dependencies: README.md
	@PYTHONPATH=. pip install packaging       ##here we pip install packaging before calling the script that needs it.
	@baselayer/tools/check_app_environment.py
	@PYTHONPATH=. python baselayer/tools/pip_install_requirements.py baselayer/requirements.txt requirements.txt
	@./baselayer/tools/silent_monitor.py baselayer/tools/check_js_deps.sh
```

Theodlz:
@stefanv any thoughts? Last thing to figure out before I can pin the latest baselayer in Skyportal.

PR #375, opened 2024-01-18
Install packaging first separately

Theodlz:
An attempt/suggestion to fix https://github.com/cesium-ml/baselayer/issues/374

ALSO: we try migrating to pre-commit.ci now that the pre-commit github action is deprecated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,7 @@ help:
 	@python ./baselayer/tools/makefile_to_help.py $(MAKEFILE_LIST)
 
 dependencies: README.md
+	@PYTHONPATH=. pip install packaging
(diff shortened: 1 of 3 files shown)


Theodlz:
the new pre-commit thingy didn't run here, maybe it needs to be merged first but I'll try separately if I can

Theodlz:
ok I got it to run there: https://github.com/Theodlz/baselayer/pull/3, but didn't work as expected I think. Changed a lot of imports that the previous pre-commit GA or the local version wouldn't have. Not sure their CI think is working as intended.

Theodlz:
removed the pre-commit changes

PR #377, opened 2024-01-23
Nginx conf: list <server> SSL only if a certificate is used

Theodlz:
The current configuration is incorrect. In the app, the `ssl` key of the config is used to figure out if the server should be accessed with https (if True) or http (if False), and not whether or not nginx should use an SSL certificate. For that, we have the `use_ssl_certificate` key already present in the default config.

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -20,8 +20,11 @@ paths:
     downloads_folder: '/tmp'
 
 server:
-    # If your server is expected to be accessed with https and not http,
-    # then set this to True
+    # Set this to true if the URL of your server is expected to be secure, i.e.
+    # https://...
+    # E.g., if you are using an HTTPS load balancer, this value will likely be true.
+    #
+    # Also see `ssl_certificate` for running nginx in SSL mode. 
     ssl: False
 
     # If you are using certificates, you can specify them here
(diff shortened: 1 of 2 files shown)


config.yaml.defaults
```diff
@@ -20,10 +20,8 @@ paths:
     downloads_folder: '/tmp'
 
 server:
-    # Whether the base URL is http or https
-    #
-    # If you are behind a load balancer that connects to the app via http,
-    # this should be set to False.
+    # If your server is expected to be accessed with https and not http,
+    # then set this to True
```
stefanv, line 24:
```suggestion
    # Set this to true if the URL of your server is expected to be secure, i.e.
    # https://...
    # E.g., if you are using an HTTPS load balancer, this value will likely be true.
    #
    # Also see `ssl_certificate` for running nginx in SSL mode. 
```

services/nginx/nginx.conf.template
```diff
@@ -111,7 +111,7 @@ http {
     {% if env.debug %}
     listen 127.0.0.1:{{ ports.app }};
     {% else %}
-    {% if server.ssl %}
+    {% if server.ssl_use_certificate %}
```
stefanv, line 114: We don't need this field, I think; `ssl_certificate` is enough by itself.

stefanv: Should be OK? Empty strings are also False.

Theodlz: I suppose so, just wondering if this could be the reason why this argument was added. But yes, it should work just fine.

PR #379, opened 2024-02-06
Brotli compression support

Theodlz:
Don't review yet, just a draft

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -76,6 +76,9 @@ jobs:
 
           sudo apt install -y wget nodejs unzip firefox
 
+          # if nginx is already installed, remove it
+          sudo apt remove -y nginx nginx-common nginx-core
+
           # install nginx from source so we can add the brotli module
           git clone --recursive https://github.com/google/ngx_brotli.git
           wget https://nginx.org/download/nginx-1.24.0.tar.gz
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -45,10 +45,7 @@ jobs:
 
       - name: Install template app
         run: |
-          git clone https://github.com/Theodlz/baselayer_template_app
-          cd baselayer_template_app
-          git checkout brotli
-          cd ..
+          git clone https://github.com/cesium-ml/baselayer_template_app
           cp -rf baselayer baselayer_template_app/
 
       - uses: actions/cache@v3
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -46,7 +46,9 @@ jobs:
       - name: Install template app
         run: |
           git clone https://github.com/Theodlz/baselayer_template_app
+          cd baselayer_template_app
           git checkout brotli
+          cd ..
           cp -rf baselayer baselayer_template_app/
 
       - uses: actions/cache@v3
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -45,7 +45,8 @@ jobs:
 
       - name: Install template app
         run: |
-          git clone https://github.com/cesium-ml/baselayer_template_app
+          git clone https://github.com/Theodlz/baselayer_template_app.git
+          git checkout brotli
           cp -rf baselayer baselayer_template_app/
 
       - uses: actions/cache@v3
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -88,7 +88,7 @@ jobs:
           export CURRENT_DIR=$(pwd)
           cd nginx-1.24.0
           ./configure --sbin-path=/usr/bin/nginx --conf-path=/usr/local/nginx/nginx.conf --pid-path=/usr/local/nginx/nginx.pid --with-http_ssl_module --with-stream --with-mail=dynamic --with-http_realip_module --with-compat --add-module=${CURRENT_DIR}/ngx_brotli
-          make && make install
+          sudo make && sudo make install
           nginx
 
           pip install --upgrade pip
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -89,7 +89,7 @@ jobs:
           cd nginx-1.24.0
           ./configure --sbin-path=/usr/bin/nginx --conf-path=/usr/local/nginx/nginx.conf --pid-path=/usr/local/nginx/nginx.pid --with-http_ssl_module --with-stream --with-mail=dynamic --with-http_realip_module --with-compat --add-module=${CURRENT_DIR}/ngx_brotli
           sudo make && sudo make install
-          nginx
+          sudo nginx
 
           pip install --upgrade pip
           pip install wheel
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -45,7 +45,7 @@ jobs:
 
       - name: Install template app
         run: |
-          git clone https://github.com/Theodlz/baselayer_template_app.git
+          git clone https://github.com/Theodlz/baselayer_template_app
           git checkout brotli
           cp -rf baselayer baselayer_template_app/
 
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -79,20 +79,10 @@ jobs:
           # if nginx is already installed, remove it
           sudo apt remove -y nginx nginx-common nginx-core
 
-          # install nginx from source so we can add the brotli module
-          git clone --recursive https://github.com/google/ngx_brotli.git
-          wget https://nginx.org/download/nginx-1.24.0.tar.gz
-          tar zxf nginx-1.24.0.tar.gz
-          cd ngx_brotli/deps/brotli
-          mkdir out && cd out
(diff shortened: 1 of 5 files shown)


tools/fill_conf_values.py
```diff
@@ -36,12 +36,35 @@ def hash_filter(string, htype):
...
 custom_filters = {"md5sum": md5sum, "version": version, "hash": hash_filter}
 
+USE_BROTLI = nginx_brotli_module_installed()
+IS_MACOS = is_macos()
+
 
 def fill_config_file_values(template_paths):
     log("Compiling configuration templates")
     env, cfg = load_env()
+    cfg["server"]["use_brotli"] = USE_BROTLI
```
stefanv, line 66: The templates become quite confusing this way. Can we use `cfg["fill-config-feature"]` or similar?

tools/fill_conf_values.py
```diff
@@ -36,12 +36,35 @@ def hash_filter(string, htype):
...
+    cfg["server"]["is_macos"] = IS_MACOS
```
stefanv, line 67: Probably not needed; just do all the logic for whether or not brotli should be loaded here.

tools/fill_conf_values.py
Theodlz: sounds good!

tools/fill_conf_values.py
Theodlz: actually it is, because on mac once installed with homebrew brotli doesnt need to be dynamically loading in the nginx config. But if installed with linux it looks like you need to. So I do need to have both of these as separate arguments

Theodlz: what I need to change is that I actually need to find the full path to the brotli module, which is trickier. But that way no need to figure out if its mac or not. I just need to find the path to where the nginx modules are stored and I'm done, I can always dynamically load brotli

Theodlz: Yep, just got it to work on linux. I just need to verify that the dynamic import on macos is possible, and then I can have one single method for both.

Theodlz: @stefanv I managed to install brotli both dynamically and statically in nginx while on linux. I think I've got it figured out. I'll add a piece of code which if statically installed, does not add the lines in the baselayer config to dynamically load brotli, and if installed dynamically, tries to figure out where the modules are so they can be loaded. If it can't figure out where the modules are but brotli is installed, just display a warning when running the app and start it without brotli

Theodlz:
@stefanv I need to make some changes to the GA (like go back to the real baselayer app template not my fork) before merging anything, but this works very well! The code snippet from the GA can be reused in the SkyPortal dockerfile to install nginx with brotli :). I will also add some documentation before merging, but re-requesting a review so you can see the new code.

.github/workflows/test.yaml
```diff
@@ -71,7 +74,25 @@ jobs:
...
+          cd ngx_brotli/deps/brotli
+          mkdir out && cd out
+          cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF -DCMAKE_C_FLAGS="-Ofast -m64 -march=native -mtune=native -flto -funroll-loops -ffunction-sections -fdata-sections -Wl,--gc-sections" -DCMAKE_CXX_FLAGS="-Ofast -m64 -march=native -mtune=native -flto -funroll-loops -ffunction-sections -fdata-sections -Wl,--gc-sections" -DCMAKE_INSTALL_PREFIX=./installed ..
+          cmake --build . --config Release --target brotlienc
+          cd ../../../..
+          export CURRENT_DIR=$(pwd)
+          cd nginx-1.24.0
+          ./configure --sbin-path=/usr/bin/nginx --conf-path=/usr/local/nginx/nginx.conf --pid-path=/usr/local/nginx/nginx.pid --with-http_ssl_module --with-stream --with-mail=dynamic --with-http_realip_module --with-compat --add-module=${CURRENT_DIR}/ngx_brotli
+          sudo make && sudo make install
+          sudo nginx
```
stefanv, line 95: It's nice to be able to test this, but maybe a bit much for an optional feature.

doc/setup.md
```diff
@@ -37,7 +38,50 @@ See [below](#configuration) for more information on modifying the baselayer conf
...
+  cd ngx_brotli/deps/brotli
+  mkdir out && cd out
+  cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF -DCMAKE_C_FLAGS="-Ofast -m64 -march=native -mtune=native -flto -funroll-loops -ffunction-sections -fdata-sections -Wl,--gc-sections" -DCMAKE_CXX_FLAGS="-Ofast -m64 -march=native -mtune=native -flto -funroll-loops -ffunction-sections -fdata-sections -Wl,--gc-sections" -DCMAKE_INSTALL_PREFIX=./installed ..
+  cmake --build . --config Release --target brotlienc
+  cd ../../../..
+  export CURRENT_DIR=$(pwd)
+  cd nginx-1.24.0
+  ./configure --sbin-path=/usr/sbin/nginx --conf-path=/usr/local/nginx/nginx.conf --pid-path=/usr/local/nginx/nginx.pid --with-http_ssl_module --with-stream --with-mail=dynamic --with-http_realip_module --with-compat --add-module=${CURRENT_DIR}/ngx_brotli
+  sudo make && sudo make install
+  ```
```
stefanv, line 60: What's wrong with https://packages.debian.org/sid/libnginx-mod-http-brotli-filter?

tools/fill_conf_values.py
```diff
@@ -36,12 +37,76 @@ def hash_filter(string, htype):
...
+    """Check if the nginx brotli module is installed
+
+    Returns
+    -------
+    installed : bool
+        True if the nginx brotli module is installed, False otherwise
+    dynamic : bool
+        True if the module is dynamically loaded, False otherwise
+    modules_dir : str
+        The directory where the nginx modules are located if dynamic loading is used
```
stefanv, line 50: Somehow, a NamedTuple feels like it might work well here, but that's total nitpicking and not even clearly an improvement :)

tools/fill_conf_values.py
```diff
@@ -36,12 +37,76 @@ def hash_filter(string, htype):
...
+        elif (
+            "--add-dynamic-module" in output
+            and "brotli" in output.split("--add-dynamic-module")[1].strip()
+        ):
+            installed = True
+            dynamic = True
+            # we try to figure out where the modules are, there are 2 possibilities
+            # 1. the --modules-path directive is used
+            # 2. the default directory is used, which is where the configuration file is located so we look for it
+            if "--modules-path" in output:
```
stefanv, line 75: nginx has a search path, so will

```
load_module modules/ngx_http_brotli_filter_module.so;
load_module ...
```

not suffice?

stefanv (commented):
Looks good! A few small questions.

.github/workflows/test.yaml
Theodlz: I agree, was just a good way for me to get it to run on github and see if it works as expected. Will remove it!

Maybe we can keep that in skyportal though, and add the test there

stefanv: Or, if the simpler Debian instructions work, that'd be fine.

tools/fill_conf_values.py
Theodlz: hahaha no of course. I'll try that if I have the time.

tools/fill_conf_values.py
Theodlz: Well I actually tried that on my linux machine, and it didn't work! Not sure why... but it did look into the directory where the config is and didn't look at any other path.

.github/workflows/test.yaml
Theodlz: Do you mean with libnginx-mod-http-brotli-filter? Well that was actually the issue for me, the lack of instruction. NGINX is supposed to be compile with the modules, so I didn't quite understand how installing this would work, nor what it did.

I can give it a go, but I found it more confusing than anything else.

stefanv: I think you just need to install this package and then you can use `load_module modules/ngx_http_brotli_filter_module.so;`.

Theodlz: if that works that's great! But then it means I can't detect whether or not brotli is installed dynamically by running `nginx -V`, which I suspect this package doesn't edit. Could work too, I'll give it a shot.

Theodlz: @stefanv a simple apt install can't actually find `libnginx-mod-http-brotli-filter`. Did you install it differently?

tools/fill_conf_values.py
stefanv: Do you have a minimal config you use to test with?

Theodlz: What do you mean? I basically run SkyPortal with that modified version of baselayer. I ran it on an ubuntu machine with 8GB of ram, that had nginx but that I replaced with these instructions.

Theodlz:
@stefanv updated the GA and the docs to install from the ppa, and updated the python logic to find the nginx brotli module. This is solid, works well on both dynamic and static installations (and on both windows and mac).

Let me know what you think. If we want this tested, we will have to update the baselayer template app with a test (like the one I added to the fork pinned here for testing). So if you are happy with these changes, let me know and I'll pin the normal template app again and we can merge :)

stefanv (approved):

stefanv:
I like it! Let's also merge the test on BL.

Theodlz:
Great! I went back to git cloning the real template app in the GA. I'll open the PR on the template app ASAP.

PR #380, opened 2024-02-23
Query accessible records: SQLAlchemy 2.0 syntax

Theodlz:
I noticed that when using the sqlachemy 2.0 syntax to create `CustomUserAccessControl`s on a table, I couldn't commit().

This is because commit() calls a method that only supported the `sa.Query` type of objects, and not the new `sa.Select` that you are expected to use in SQLAlchemy 2.0

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -74,11 +74,13 @@ jobs:
           sudo apt install -y wget nodejs unzip firefox
 
           # if nginx is already installed, remove it
-          sudo apt remove -y nginx nginx-common nginx-core
+          sudo apt remove -y nginx nginx-common nginx-core nginx-full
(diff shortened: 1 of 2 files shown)


stefanv:
Great detective work! Do you have a sense of how we can test this?

Theodlz:
> Great detective work! Do you have a sense of how we can test this?

it’s a little tricky to test for sure. For that we would need one or 2 model(s) so that we can have 2 custom accessibility rules on update: one using DBSession.query() and one using sa.select().

In SP most of these rules are outdated and use DBSession and not sa.select(). So as long as what works still works and what is now written with sa.select() works too, we are covered.

But I 100% agree that some kind of dedicated test could be good.

Maybe we can actually test this without using these accessibility rules at all, just emulate whatever would be the input of these 2 methods I edited to take in both kind of inputs. Let me see if I can figure something out tomorrow.

PR #382, opened 2024-03-28
Websocket Frontend Notifications: specify duration

Theodlz:
Users have often complained that some notifications were too hard to read because of the short 3 seconds we have by default. We might want to change that default altogether, but it seems like a good idea to let the developer specify a longer duration from the backend if needed (when sending the websocket notification), depending on how long the `note` that will be displayed is.

This PR modifies the existing `notificationHandler` to try and grab a duration from the websocket payload, and set it to the default 3 seconds if:
- it is missing
- it is equal or lower than 0
- it is higher than 30 seconds


mcoughlin (approved):
LGTM

PR #383, opened 2024-05-07
User: bio and bot flag

Theodlz:
The idea of this PR is to add a bio field to the user model, as well as a bot flag (maybe should be renamed to `is_bot` to make the boolean-ness more obvious where we use it???).

it's nice in general to have a field for a user to store a bio that other users could access, and this will come in handy for a couple of applications:
- we're considering a public user profile page, where other users that are in any of your groups (except public of course, as all users are there) can see your user profile's most basic info, as well as some kind of activity tracker similar to the news feed but only for a given user.
- for the TNS reporting, before we can let bot accounts report to the Transient Name Server autonomously (needed for BTSbot), we need to be able to mark bot accounts as such (hence the new flag), but also have some kind of remark added to the TNS report, which would be that `bio` field for a bot account.

Naturally, this will be followed by a SkyPortal PR which code is already written (the PR will be opened once this is merge to main).

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1876,7 +1876,7 @@ class User(Base):
     preferences = sa.Column(
         JSONB, nullable=True, doc="The user's application settings."
     )
-    bot = sa.Column(
+    is_bot = sa.Column(
         sa.Boolean,
         nullable=False,
         server_default="false",
--- a/app/models.py
+++ b/app/models.py
@@ -1851,6 +1851,11 @@ class User(Base):
 
     first_name = sa.Column(sa.String, nullable=True, doc="The User's first name.")
     last_name = sa.Column(sa.String, nullable=True, doc="The User's last name.")
+    bio = sa.Column(
+        sa.String,
+        nullable=True,
+        doc="A short biography of the user, or description for bot accounts.",
+    )
     affiliations = sa.Column(
         sa.ARRAY(sa.String),
         nullable=False,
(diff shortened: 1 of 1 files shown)


stefanv:
@Theodlz The `is_bot` field feels quite specific; should we add a `JSONB` column so you can add anything in there you want?

Theodlz:
@stefanv the idea is that we really need a way to differentiate real user accounts, and bot accounts (that are used by scripts). On the contrary, I feel like having that in a JSONB doest make it obvious enough that this is not a normal account but a bot.

Does that make sense?

Theodlz:
But I do agree that a JSONB is nice and flexible, and avoid opening other PRs in the future. In which case yes it could be a nice idea.

stefanv:
I'm not sure I understand. I was thinking that baselayer is generic, and `is_bot` is a specialized request, meaning it may be better to have a JSONB column and to implement that functionality directly in SkyPortal.

Theodlz:
Ok, I understand @stefanv . I think that we just have a different understand/view of this feature. I would argue that even the generic baselayer core would benefit from a way to distinguish API-only accounts from real users, but I might be wrong there.

Ultimately both options are really fine by me, so I'm happy to replace this by a JSONB column. Any suggestions for that column's name? Could be something like `details`?

stefanv:
Probably not important enough to hold this PR up; we can also just find a more descriptive name of the field for now. How about `is_human`? Systems can use that for captcha or other verification as well.

stefanv:
Let's rename to `is_bot` and merge.

Theodlz:
done @stefanv.

PR #384, opened 2024-06-10
check for geckodriver binary before bothering to install it

mcoughlin:
This PR checks for geckodriver binary before bothering to install it and avoids some requests that seem to be overwhelming the GA.


stefanv:
Thanks!

PR #385, opened 2024-06-28
Document debug mode

stefanv:
Not sure it's in the right spot, but it should be *somewhere*.


Theodlz (approved):
LGTM 🚀 Definitely good to have that here.

PR #387, opened 2024-09-19
auth_or_token decorator: set __authenticated__=True attribute

Theodlz:
add an __authenticated__=True attribute to handler methods that require auth, so we can use it when autogenerating the openapi specs and show/hide token auth headers in the auto-generated code examples.

Required for us to then merge https://github.com/skyportal/skyportal/pull/5273

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -14,7 +14,7 @@ jobs:
     if: github.repository_owner == 'cesium-ml'
 
     steps:
-      - uses: actions/cache@v3
+      - uses: actions/cache@v4
         with:
           path: |
             ~/.cache/pip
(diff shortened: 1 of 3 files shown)


stefanv:
We'll need to fix the BL tests using importlib.metadata and
importlib.resources, instead of pkg_resources.

Theodlz:
> We'll need to fix the BL tests using importlib.metadata and importlib.resources, instead of pkg_resources.

perfect timing! I actually reverted that python version upgrade just for the sake of this PR, and was reading through a few articles to try to figuring out the pkg_resources replacement situation

Theodlz:
didn't exactly revert it, but downgraded to 3.10

PR #388, opened 2024-09-30
webpack -> rspack migration

Theodlz:
Bundling times for SkyPortal (in production mode for example) are cut in half using rspack, and it looks like the bundle sizes are also slightly reduced in production mode.

Overall rspack seems to offer us the same features as webpack, while speeding up things which is very handy while developing, and while running any CI/CD.

I'd be in favor of moving to rspack, especially given how minimal the changes needed are.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -45,8 +45,7 @@ jobs:
 
       - name: Install template app (modified, temporary)
         run: |
-          git clone https://github.com/Theodlz/baselayer_template_app
-          git checkout rspack-v1
+          git clone -b rspack-v1 https://github.com/Theodlz/baselayer_template_app
           cp -rf baselayer baselayer_template_app/
 
       - uses: actions/cache@v4
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -34,9 +34,9 @@ jobs:
     steps:
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.8"
+          python-version: "3.10"
 
-      - uses: actions/setup-node@v3
+      - uses: actions/setup-node@v4
         with:
           node-version: 20
 
@@ -48,13 +48,13 @@ jobs:
           git clone https://github.com/cesium-ml/baselayer_template_app
           cp -rf baselayer baselayer_template_app/
 
-      - uses: actions/cache@v3
+      - uses: actions/cache@v4
         with:
           path: |
             ~/.npm
           key: ${{ runner.os }}-npm-${{ hashFiles('baselayer_template_app/package.json') }}
 
-      - uses: actions/cache@v3
+      - uses: actions/cache@v4
         with:
           path: |
             ~/.cache/pip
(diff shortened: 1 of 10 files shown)


stefanv:
If it's a 1-to-1 replacement, that's fine; we have a somewhat complex webpack configuration, so I wanted to make sure that it takes all behavior into consideration.

Theodlz:
@stefanv I had to edit the CI to point to a custom branch of the template app (that will be its own PR), but CI is running and everything is working as expected.

I checked the config, and `rspack` runs some pretty strict validation, which should make it error out if the config contains things that are not supported, instead of ignoring it. So, this should mean that our config is valid. I looked through the rest of the doc, and none of our current config isn't supported. You can see that here: https://rspack.dev/guide/migration/webpack#updating-configuration

Theodlz:
let me know when/if you are happy with this, and I'll remove the temporary changes, and we can look at the PR on the template app.

Theodlz:
https://github.com/cesium-ml/baselayer_template_app/pull/304 Template App PR is ready

https://github.com/skyportal/skyportal/pull/5278 SkyPortal PR is ready, can be merged once baselayer's is.

stefanv:
I'm happy to give this a try. Correct order seems to be: baselayer template app, then updated version of this, then skyportal.

I think it would be good to have a note somewhere that we use rspack as a 1:1 replacement of webpack, in case other projects prefer to use that.

Theodlz:
> I'm happy to give this a try. Correct order seems to be: baselayer template app, then updated version of this, then skyportal.
> 
> I think it would be good to have a note somewhere that we use rspack as a 1:1 replacement of webpack, in case other projects prefer to use that.

That's a great idea. Would you rather have me add this to baselayer, or the template app that has the rspack.config.js?

stefanv:
Baselayer docs, maybe, since that's where the command is being invoked from?

Theodlz:
@stefanv would you rather have me merge baselayer pointing to the edited template app (for the CI to be green on main), then merge the PR on the template app, then open another one on baselayer to go back to the main of the template app?

Or, I get this PR to go back to the main branch of the template app, we merge it and CI fails, but once the template app's PR is merged we just rerun the CI here to have it be green.

Any preference? I feel like the second option is best, since I just need to rerun an action rather than re-open a PR.

Theodlz:
adding the docs changes in a minute btw

Theodlz:
done @stefanv, I added a short paragraph/section dedicated to the web app bundling microservice

Theodlz:
> Baselayer docs, maybe, since that's where the command is being invoked from?

done, I added a short paragraph/section dedicated to the web app bundling microservice

.github/workflows/test.yaml
```diff
@@ -43,9 +43,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Install template app
+      - name: Install template app (modified, temporary)
         run: |
-          git clone https://github.com/cesium-ml/baselayer_template_app
+          git clone -b rspack-v1 https://github.com/Theodlz/baselayer_template_app
```
stefanv, line 48: Can merge as soon as this has returned to cesium-ml

stefanv:
Thanks!

PR #396, opened 2025-06-15
Add group accessible function for User and Token

antoine-le-calloch:
This PR introduces the AccessVerificationMixin to centralize access checks for both User and Token models.
It also adds a reusable assert_group_accessible method to easily check whether a user (or token) can access a given group, which are performed frequently throughout SkyPortal. This refactor helps clean up redundant skyportal codes, and for me it makes sense to put it here.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1946,6 +1946,23 @@ class User(Base):
         """The base model for User subclasses."""
         return User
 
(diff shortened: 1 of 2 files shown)


antoine-le-calloch:
@Theodlz @mcoughlin @stefanv (Sorry can't add reviewers)

PR #412, opened 2026-03-09
Extensions: clone with --revision - fallback 

Theodlz:
add fallback to clone without --revision, if that flag isn't available for the version of `git` installed on the host (we ran into that issue in prod, where even recent versions of ubuntu, debian, ... have a version of `git` that's a little old and does not support this flag, while our dev machines on macos do hence why we never saw this commit).

I could have simply dropped the `--revision` flag altogether (since anyway we checkout to the right revision if need be, few lines after), but I like the idea of only copying over the exact revision needed when we first clone the repo, which is a bit less wasteful than clone + checkout. So, might as well try it and only if it doesn't work take the regular clone + checkout path (+ log that we ran into that fallback and why, to give folks a hint about their git version perhaps being a little old).


mcoughlin (approved):
LGTM

stefanv (approved):

PR #413, opened 2026-03-09
JS deps .sh scripts (updates & check), always use uv's python

Theodlz:
In this PR, we update 2 .sh scripts to use `PYTHONPATH=. uv run python` rather than just `python`, so it works whether or not one has entered the virtual env that was created by uv. I found that change quite useful because we that implemented, I can now just run `make run` in SkyPortal and my app starts, and I don't even need to think about entering my venv first!

### DIFF
--- a/tools/check_js_deps.sh
+++ b/tools/check_js_deps.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-INSTALLER=$(python baselayer/tools/get_js_installer.py)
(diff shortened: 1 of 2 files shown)


mcoughlin (approved):
LGTM

Theodlz:
> Is the `PYTHONPATH=.` part necessary here? Those scripts run without any baselayer dependency, I think.

Might not be! I thought of adding it anyways in case we do have some dependencies at some point, but happy to remove it if you think it can cause issues!

PR #416, opened 2026-03-12
[WIP] psycopg2 -> psycopg

Theodlz:
`psycopg2` isn't maintained anymore (at least no updates since Oct 2025). As its readme says on pypi, now things have moved to the `psycopg` package (I know, the numbering system sucks, it's like battlefield one all over again). So, this PR is just a quick attempt at pinning that new package rather than the old one.

For context, I've been working on a SkyPortal PR that updates dependencies, and trying to get numpy v2 to work there. Turns out that it's pretty incompatible with the current version of our sqlalchemy+psycopg2 combo, as np.float64() make their way into the SQL statement which obviously yields errors. This issue https://github.com/sqlalchemy/sqlalchemy/discussions/11712 hints at upgrading psycopg2 to psycopg to fix this. 

PS: new psycopg3 now comes with another package called `psycopg_pool`. That can be installed separately, and used in place of sqlalchemy's built in connection pool. Unclear to me if one of these 2 performs better, but the sqlalchemy docs seem to say good things about it: https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#using-psycopg-connection-pooling. Might want to consider using this now.


Theodlz:
closing in favor of #421

PR #418, opened 2026-03-26
Improve makefile bundle management

antoine-le-calloch:
- Remove hardcoded bundle filename from Makefile to allow having content hashes in the main bundle name.
- Update frontend test to remove the hardcoded bundle name.
- simplify bundle targets

I have opened a PR (https://github.com/skyportal/skyportal/pull/6034) on SkyPortal to switch the bundle cache key to use a content hash instead of the skyportal version hash. As part of this, I also updated Baselayer to support hashed bundle filenames.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,6 @@ B=\033[1m
 # Normal
 N=\033[0m
 
-bundle = static/build/main.bundle.js
 rspack = npx rspack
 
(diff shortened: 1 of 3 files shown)


Theodlz (approved):
LGTM, just one minor question

stefanv:
Sorry folks, I was out on leave.

stefanv:
Changes look good though, thanks!

PR #419, opened 2026-04-05
detect numpy scalars

mcoughlin:
This PR adds numpy scalars to the data type check (helpful for Numpy 2 transition in SkyPortal).

### DIFF
--- a/app/json_util.py
+++ b/app/json_util.py
@@ -36,6 +36,9 @@ class Encoder(json.JSONEncoder):
         elif type(o).__name__ == "ndarray":  # avoid numpy import
             return o.tolist()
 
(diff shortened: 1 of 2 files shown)


Theodlz (approved):

app/json_util.py
```diff
@@ -36,6 +36,9 @@ def default(self, o):
         elif type(o).__name__ == "ndarray":  # avoid numpy import
             return o.tolist()
 
+        elif hasattr(o, "item"):  # numpy scalar (int64, float64, bool_, etc.)
```
stefanv:
```suggestion
        elif type(o).__module__ == "numpy" and hasattr(o, "item"):  # numpy scalar
```

This feels very broad; should we constraint to numpy? All sorts of other objects have `.item`.

stefanv (approved):

PR #420, opened 2026-05-12
psycopg2 -> psycopg3

mcoughlin:
This PR begins the transition to psycopg3, which will also enable Async sessions in skyportal.


mcoughlin:
@antoine-le-calloch

Theodlz:
gentle reminder to test this along with skyportal. I already had https://github.com/cesium-ml/baselayer/pull/416 open for this, and clearly when using it in skyportal it sh*t the bed. But maybe that PR of yours already addresses that. Anyways, just a comment about testing this out with skyportal before committing anything to baselayer main. We had baselayer main in a "skyportal-incompatible" state for months at some point, and that made adding any new baselayer-based feature complicated.

PR #421, opened 2026-05-17
psycopg3

mcoughlin:
This PR only focuses on the psycopg2 -> psycopg3 changes (async will come on a different branch).


app/models.py
```diff
@@ -266,8 +267,6 @@ def init_db(
     }
     conn = sa.create_engine(
         url,
-        client_encoding="utf8",
-        executemany_mode="values_plus_batch",
```
Theodlz, line 270: might be good to check how it now behaves in psycopg3, where that parameter doesn't exist. Is there another parameter we need to set to get the same behavior, do we not need that anymore, etc?

mcoughlin: Yeah looks both of these were psycopg2 only. PostgreSQL defaults to UTF-8 so should be good there:
    >>> conn.execute(text("SHOW client_encoding")).scalar()
insertmanyvalues is the replacement for  executemany_mode, and defaults to true:
    >>> PGDialect_psycopg().use_insertmanyvalues
       True
   >>> PGDialect_psycopg().insertmanyvalues_page_size
        1000

Theodlz (approved):
LGTM

PR #422, opened 2026-05-17
Async

mcoughlin:
async capabilities on top of https://github.com/cesium-ml/baselayer/pull/421


pyproject.toml
```diff
@@ -9,7 +9,8 @@ dependencies = [
     "numpy>=1.21.4",
     "packaging>=23.0",
     "phonenumbers>=8.12.15",
-    "psycopg2-binary>=2.8.6",
+    "psycopg[binary]>=3.2.0",
+    "greenlet>=3.0.0",
```
Theodlz: what does `greenlet` do? Asking because I see no explicit import to it here.

mcoughlin: It is the dependency for the async stuff. Could also import this as sqlalchemy[asyncio].

Theodlz: yes that's definitely a lot clearer. Thanks

app/access.py
```diff
@@ -1,11 +1,31 @@
...
 from baselayer.app.custom_exceptions import AccessError  # noqa: F401
-from baselayer.app.models import DBSession, Role, Token, User  # noqa: F401
+from baselayer.app.models import (  # noqa: F401
+    DBSession,
+    Role,
+    Token,
+    User,
+    async_plain_session_factory,
+)
+
```
antoine-le-calloch, line 15: `async_plain_session_factory` can be remove here.

antoine-le-calloch (approved):
LGTM, one small comment.

app/access.py
mcoughlin: Fixed!

antoine-le-calloch:
@mcoughlin another comment is that the sync stuff pass `insertmanyvalues_page_size=EXECUTEMANY_PAGESIZE` but not the async.

mcoughlin:
@antoine-le-calloch Got that one too, thanks!

PR #423, opened 2026-05-31
latency logging

mcoughlin:
This PR adds latency logging to the nginx logs (so we can measure per-endpoint latencies).


stefanv:
Thanks; we can also consider moving to structured logs soon if that would help accommodate more fields.

PR #424, opened 2026-06-04
Allow the websocket message bus to bind a tcp:// address

mcoughlin:
This PR allows the websocket message bus to bind a tcp:// address (next step in Kubernetes readiness).

### DIFF
--- a/app/flow.py
+++ b/app/flow.py
@@ -11,11 +11,15 @@ log = make_log("flow")
 class Flow:
     """Send messages through websocket to frontend"""
 
-    def __init__(self, socket_path=cfg["ports.websocket_path_in"]):
-        self._socket_path = socket_path
+    def __init__(
+        self,
+        host=cfg["hosts.message_bus"],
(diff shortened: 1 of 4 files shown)


stefanv:
Wondering why we don't just standardize on using ports internally too, then we can avoid the munging.

Note we have two servers here: message bus + websocket server. The websocket server is, in fact, just a client of the message bus.

The message bus should be straightforward: runs a single instance, and exposes two ports — one to receive messages, one to publish messages.

The websocket server connects to the port where messages are published, and forwards them to the browser.

The app connects to the port where messages are received, and pushes them onto the bus.

So, let's just use TCP for everything—simpler?

```yaml
ports:
  websocket: ...
  message_bus_receive: ...
  message_bus_publish: ...
hosts:
  message_bus: 127.0.0.1
```

Flow will push to `hosts.message_bus`:`message_bus_receive`.

stefanv:
Thanks!

PR #425, opened 2026-06-11
anonymous login

mcoughlin:
This PR enables anonymous login (if specified in the config).

### DIFF
--- a/app/access.py
+++ b/app/access.py
@@ -48,12 +48,12 @@ def auth_or_token(method):
             if self.current_user is not None:
                 if not self.current_user.is_active():
                     raise tornado.web.HTTPError(403, "User account expired")
-                # Anonymous access serves a user without a signed-in `user_id`
-                # cookie; restrict it to safe (read-only) HTTP methods.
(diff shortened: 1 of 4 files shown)


mcoughlin:

antoine-le-calloch:
@mcoughlin Nothing is blocking a "read-only" user from using write HTTP endpoints (POST/PUT).

mcoughlin:
@antoine-le-calloch great point. fixed.

antoine-le-calloch:
@mcoughlin Anonymous read-only is enforced by checking for a missing user_id cookie, but get_current_user() also falls back to the anonymous account when a user_id cookie is present but invalid. So those requests keep write access instead of being restricted to read-only.

mcoughlin:
@antoine-le-calloch Thanks! One more look?

antoine-le-calloch:
@mcoughlin LGTM. Maybe one last thing would be to add it to the default config and set it to false, so people can see from the .default file that this key exists.

PR #428, opened 2026-06-20
remove notedown

mcoughlin:
This PR removes the unused notedown package (which pulls in the deprecated bleach package).


mcoughlin:
@leonitousconforti @thomasculino

PR #430, opened 2026-07-04
Connection pooling

mcoughlin:
This PR introduces an optional transaction-mode connection pooler.


mcoughlin:

antoine-le-calloch:
@mcoughlin LGTM

PR #431, opened 2026-07-04
opt out of created_at index

mcoughlin:
This PR enables better index usage, eg opt out of an index on created_at opt-out and instead use a composite index.


mcoughlin:

mcoughlin:

leonitousconforti (approved):

PR #432, opened 2026-07-05
fix permissions on OR-composed read control with a column-filtered custom generator

mcoughlin:
This PR fixes a latent bug where if you had a OR-composed read control with a column-filtered custom generator, verified through the 2.0 select path, it would over grant permissions (we did not have such a thing before some of the stream work I am doing).

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -1378,9 +1378,15 @@ class CustomUserAccessControl(UserAccessControl):
         else:
             stmt = self.query_generator(cls, user_or_token)
 
-        # retrieve specified columns if requested
+        # retrieve specified columns if requested. Project the subquery's own
+        # columns rather than `columns` (which reference the outer mapped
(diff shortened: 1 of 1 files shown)


app/models.py
```diff
@@ -1378,9 +1378,15 @@ def select_accessible_rows(self, cls, user_or_token, columns=None):
         else:
             stmt = self.query_generator(cls, user_or_token)
 
-        # retrieve specified columns if requested
+        # retrieve specified columns if requested. Project the subquery's own
```
stefanv, line 1381: I don't follow this comment, so I'll ask @antoine-le-calloch if he feels comfortable merging.

The fix I get, roughly, and it should go in if it closes a security hole.

antoine-le-calloch: LGTM now.

antoine-le-calloch (approved):
LGTM

PR #434, opened 2026-07-06
_resolve_pooler: force prepare_threshold=None over caller connect_args

mcoughlin:
This PR forces prepare_threshold=None over caller connect_args for _resolve_pooler.


mcoughlin:
@antoine-le-calloch

antoine-le-calloch:
> @antoine-le-calloch 

LGTM

PR #436, opened 2026-07-12
Make access-control checks primary-key-agnostic to support composite PKs

mcoughlin:
This PR makes access-control checks primary-key-agnostic to support composite PKs.


mcoughlin:
@leonitousconforti @thomasculino

leonitousconforti (approved):
LGTM

PR #438, opened 2026-07-13
Update `log.py`

leonitousconforti:
On every single log message we were recalculating 

`color_table[zlib.crc32(app.encode("ascii")) % len(color_table)]`

but appName never changes so that was redundant work on every message. It should be calculated once

### DIFF
--- a/log.py
+++ b/log.py
@@ -73,8 +73,8 @@ def __colorize(
 
 
 def make_log(appName: str) -> Callable[[str], None]:
-    color_table = ["red", "green", "yellow", "blue", "magenta", "cyan", "white"]
-    color = color_table[hash(appName.encode("ascii")) % len(color_table)]
+    app_color_table = ["red", "green", "yellow", "blue", "magenta", "cyan", "white"]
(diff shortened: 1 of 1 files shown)


log.py
```diff
@@ -29,55 +40,46 @@ def colorize(s, fg=None, bg=None, bold=False, underline=False, reverse=False):
...
-    reverse : bool
-        Whether or not to show the text in reverse video.
+    # Foreground colors go from 30-39
+    if foreground in COLOR_TABLE:
+        style_fragments.append(COLOR_TABLE.index(foreground) + 30)
 
-    Returns
-    -------
-    A string with embedded color escape sequences.
-    """
```
leonitousconforti, line 51: Thoughts on getting rid of some comments like these since IDE hover tools could be the source of truth and these comments can easily get out of date?

Image

stefanv: Type annotations tell you about the types, but there's no other place to get a description of what the parameters are, how they are to be used, etc.

The type annotation for fg here should probably be a literal with the options available.

PR #439, opened 2026-07-14
enable templates for microservices

mcoughlin:
This PR enables templates for microservices.


mcoughlin:
@thomasculino @leonitousconforti @antoine-le-calloch

stefanv:
Makes sense! Ensure that requirements are updated, also for template app.

stefanv (approved):

mcoughlin:
@stefanv jinga2 is already a baselayer dependency, and the template app inherits from baselayer, so should be good!

PR #440, opened 2026-07-15
error messages on DB failure

mcoughlin:
This PR more gracefully manages error messages on DB failure (not exposing SQL/tokens).

### DIFF
--- a/app/access.py
+++ b/app/access.py
@@ -1,5 +1,6 @@
 import functools
 import inspect
+from contextlib import contextmanager
 
 import sqlalchemy as sa
 import tornado.web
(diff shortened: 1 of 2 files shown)


mcoughlin:
@antoine-le-calloch

antoine-le-calloch:
@mcoughlin:
- login_user still uses the pre-fix pattern: on a dead connection, its except block's session.rollback() can re-raise a SQLAlchemyError that escapes the with and surfaces as a raw-SQL 500.
- In the cookie path, the inner except Exception swallows query failures and returns None, so a real DB outage during the SELECT surfaces as a misleading 401 instead of the retryable 503

mcoughlin:
Thanks @antoine-le-calloch, let me know if anything else.

antoine-le-calloch:
@mcoughlin Maybe we could extract a decorator to avoid duplicating the same code in all four places.

antoine-le-calloch:
@mcoughlin LGTM

PR #441, opened 2026-07-16
Add --test-only argument to db_init.py for test database operations.

antoine-le-calloch:


antoine-le-calloch:
@mcoughlin

mcoughlin (approved):
Ah this is a nice one.

PR #443, opened 2026-07-22
use signed in user for auth (not current as can be the anonymous viewer)

mcoughlin:
This PR uses the signed in user for auth (not current as can be the anonymous viewer).


mcoughlin:
@thomasculino @leonitousconforti

thomasculino (approved):
LGTM

PR #445, opened 2026-07-27
Share auth logic between sync/async paths; fix anonymous write bypass

stefanv:
The anonymous read-only guard lived only in the sync auth_or_token wrapper, so anonymous users could POST/PUT/DELETE to async handlers. Factor the shared authorization steps into helpers used by both paths, leaving only the token-lookup engine (sync vs async) to differ.

Addition by @antoine-le-calloch:
- Strip only the leading `token ` prefix from the `Authorization` header, so a token id containing the word `token` is no longer mangled.
- Refuse an expired account before adopting its token as the request's credentials.
- Make `is_anonymous_user` a `PSABaseHandler` attribute instead of a `getattr` default, so the anonymous read-only guard cannot fail open.
- Answer 503 rather than 500 when the anonymous user lookup hits a database failure.
- Log a blocked anonymous write as an expected error instead of an uncaught exception with a stack trace.
- Drop the `tornado.web.authenticated` call from both wrappers: the cookie path is only reached after the current-user check already raised 401, so the guard could never fire.
- Move the token lookup into `_lookup_token` / `_lookup_token_async`, leaving the sync and async wrappers identical apart from the lookup they call.
- Clean up some comments that only restate the code.


stefanv:
@mcoughlin Attempts to block the anonymous user pass-through. Needs a bit of polishing, perhaps, but the general gist.

stefanv:
@antoine-le-calloch I think this security hole still exists; hopefully this version of baselayer isn't being deployed!

antoine-le-calloch:
@stefanv I made some improvement/updates to this PR, and I think it’s ready to go on my side. I also added my additions to the PR description.

antoine-le-calloch (approved):
LGTM.

app/access.py
```diff
@@ -43,133 +40,104 @@ def _token_select_stmt(token_id):
...
+    # Reading current_user resolves the anonymous fallback and sets is_anonymous_user.
+    user = handler.current_user
+    if user is None:
+        raise tornado.web.HTTPError(
+            401,
+            'Credentials malformed; expected form "Authorization: token abc123"',
+        )
+    if not user.is_active():
+        raise tornado.web.HTTPError(403, "User account expired")
+    if handler.is_anonymous_user and handler.request.method not in SAFE_METHODS:
```
stefanv, line 81: Will this statement be reached? Wouldn't `user` be None in this case?

stefanv (commented):
Thanks, @antoine-le-calloch. This is much cleaner. I only noticed one spot where I wasn't sure of the logic. If you've tested this with SkyPortal, and all works well (and, importantly, doesn't let anonymous users through where they shouldn't!), then I think we can go ahead and merge.

app/access.py
antoine-le-calloch: User is the real "View only" account: when nobody is signed in and app.anonymous_access is on, get_current_user returns that User row and sets is_anonymous_user = True (app/handlers/base.py:65-71). user is only None when anonymous access is off or the account is missing, and then is_anonymous_user stays False, so the 401 above fires.

antoine-le-calloch:
@stefanv this have been tested with Skyportal.

PR #449, opened 2026-08-04
Add new_session for database work outside a web request

mcoughlin:
~~This PR adds a get_db_engine helper function.~~

### New version from @antoine-le-calloch :

- `new_session()`: a session of its own, for code running outside a web request.
  It takes the engine and settings `init_db()` configured, and says so clearly if
  `init_db()` has not run yet.
- `db_engine()`: the engine `DBSession` is bound to. `app/model_util.py` and the
  connection check in `app/handlers/base.py` go through it, so only one place in
  baselayer still knows where the engine is kept.
- Fixes `VerifiedSession()`, which returned a session bound to nothing:
  `_VerifiedSession.__init__` dropped every argument it was given. It only worked
  inside a handler, which set the bind by hand afterwards. That patch is gone.
- `doc/dev.md`: a new section on which session to use inside a request, outside a
  request, and when a write has to be checked against a user's permissions. Also
  renames "RLS" to "access-control" in the async comments, following review.

Used by skyportal/skyportal#6470, which replaces 30 hand-written session setups
with `new_session()`.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -45,6 +45,16 @@ session_context_id = contextvars.ContextVar("request_id", default=None)
 DBSession = scoped_session(sessionmaker(), scopefunc=session_context_id.get)
 
 
+def get_db_engine():
+    """The SQLAlchemy engine that DBSession is bound to.
+
+    A readable accessor so callers that need a one-off sync Session can write
+    ``Session(bind=get_db_engine())`` instead of reaching into
+    ``DBSession.session_factory`` internals. Only valid after ``init_db``.
+    """
+    return DBSession.session_factory.kw["bind"]
+
+
 class _VerifiedSession(sa.orm.session.Session):
     """
     Create an instance of Session when you
--- a/app/models.py
+++ b/app/models.py
@@ -45,16 +45,6 @@ session_context_id = contextvars.ContextVar("request_id", default=None)
 DBSession = scoped_session(sessionmaker(), scopefunc=session_context_id.get)
 
 
-def get_db_engine():
-    """The SQLAlchemy engine that DBSession is bound to.
-
-    A readable accessor so callers that need a one-off sync Session can write
-    ``Session(bind=get_db_engine())`` instead of reaching into
-    ``DBSession.session_factory`` internals. Only valid after ``init_db``.
-    """
-    return DBSession.session_factory.kw["bind"]
-
-
 class _VerifiedSession(sa.orm.session.Session):
     """
     Create an instance of Session when you
@@ -209,6 +199,9 @@ async_engine = None
 async_session_factory = None
 # Plain factory (no RLS check), parallel to DBSession.
 async_plain_session_factory = None
+# Sync analog: a fresh unscoped Session bound to the engine, for work that must
+# not share the request-scoped DBSession. Use ``with plain_session_factory() as s:``.
+plain_session_factory = None
 
 
 class _AsyncVerifiedSession(SAAsyncSession):
@@ -434,6 +427,9 @@ def init_db(
 
     DBSession.configure(bind=conn, autoflush=autoflush, future=True)
     Base.metadata.bind = conn
+
+    global plain_session_factory
+    plain_session_factory = sessionmaker(bind=conn, autoflush=autoflush, future=True)
 
     global async_engine, async_session_factory, async_plain_session_factory
     async_engine = create_async_engine(
--- a/app/models.py
+++ b/app/models.py
@@ -195,9 +195,10 @@ def bulk_verify(mode, collection, accessor):
 # Configured by init_db(). Sync engine/session remain authoritative for the
 # rest of the codebase; async path is opt-in per handler.
 async_engine = None
-# Verified factory (RLS check on commit), parallel to VerifiedSession.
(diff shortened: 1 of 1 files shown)


app/models.py
```diff
@@ -45,6 +45,16 @@
 DBSession = scoped_session(sessionmaker(), scopefunc=session_context_id.get)
 
 
+def get_db_engine():
+    """The SQLAlchemy engine that DBSession is bound to.
+
+    A readable accessor so callers that need a one-off sync Session can write
+    ``Session(bind=get_db_engine())`` instead of reaching into
```
stefanv: Looks good to me. Quick q: if this really is the intent, then should we just have a utility function for providing the session?

mcoughlin: good point :)

app/models.py
stefanv: I didn't see this come in: what is RLS?

app/models.py
```diff
@@ -425,6 +428,9 @@ def init_db(
     DBSession.configure(bind=conn, autoflush=autoflush, future=True)
     Base.metadata.bind = conn
 
+    global plain_session_factory
```
stefanv, line 432: LGTM; any race conditions to check for?

app/models.py
mcoughlin: just access control.

app/models.py
mcoughlin: each session should be independent, so I don't think so.

mcoughlin:
@stefanv feel free to merge if you are happy with this.

stefanv (commented):
We need some lightweight docs for how these will be used in practice.

antoine-le-calloch (approved):
I made some fix/improvement. LGTM now.

app/models.py
```diff
@@ -134,11 +134,32 @@ def commit(self):
...
         scopefunc=session_context_id.get,
     )()
 
 
+def db_engine():
+    """The engine that `init_db()` bound `DBSession` to, or None before it runs."""
+    return DBSession.session_factory.kw["bind"]
+
+
+def new_session():
```
stefanv: @antoine-le-calloch Do you think `new_session` is indicative enough of what it does? I mean, `DBSession` is also a new session on a handler. Do we perhaps need a hint such as `independent_session` or `session_no_acl` etc.?

doc/dev.md
```diff
@@ -145,6 +145,51 @@ with VerifiedSession(user_or_token) as session:
...
+from baselayer.app.models import new_session
+
+with new_session() as session:
+    session.add(record)
+    session.commit()
+```
+
+`new_session()` applies no access-control check, so it suits code acting on its own
+behalf; where a write has to be checked against a user's permissions, use
+`VerifiedSession(user_or_token)`.
```
stefanv, line 175: Hrm, but above you explained that these are scoped to `session_context_id`, so wouldn't this then overlap with other sessions?

doc/dev.md
```diff
@@ -145,6 +145,51 @@ with VerifiedSession(user_or_token) as session:
...
+
+The async factories are the counterparts of `DBSession` and `VerifiedSession`, opt-in per
```
stefanv, line 177: I think we can be clearer here about what names are the counterparts of `DBSession` and `VerifiedSession` respectively. `plain` doesn't tell the user much.

doc/dev.md
```diff
@@ -145,6 +145,51 @@ with VerifiedSession(user_or_token) as session:
...
+```
+from baselayer.app import models
+
+async with models.async_plain_session_factory() as session:
+    ...
+```
+
+Reach the async factories through the module rather than importing their names directly:
+`init_db()` rebinds those globals, so `from baselayer.app.models import
+async_plain_session_factory` captures the `None` they hold before that call.
```
stefanv, line 191: AI gobbledygook; we need to tidy this up.

doc/dev.md
```diff
@@ -145,6 +145,51 @@ with VerifiedSession(user_or_token) as session:
...
+`VerifiedSession(self.current_user)` and checks write permissions on commit, or
+`DBSession()` where that check is unwanted. Both are scoped on `session_context_id`, a
+context variable that `BaseHandler.prepare()` sets to a fresh value on every request, so
+each request works against its own session.
+
+Outside a request that context variable keeps its default of `None`, so every
+`DBSession()` call in the process resolves to one shared session; in a microservice, a
+script, or work handed off to a thread, that sharing means one caller's `rollback()`
+discards another caller's pending work. Use `new_session()` there, which builds an
+independent session on each call:
```
stefanv, line 163: In this case, I think the technical explanation is perhaps warranted, because it may be the easiest way to explain the concept. However, in general, it is better for readers when we describe how the system works (our API) instead of telling them how things are implemented under the hood—unless that enhances understanding.

Claude and other LLMs will, unless otherwise instructed, include way too much of this level of detail.

stefanv (commented):
One real concern in the docs there; the others are documentation preferences.

antoine-le-calloch:
@stefanv done in https://github.com/cesium-ml/baselayer/pull/467 !

stefanv:
Thank you, @antoine-le-calloch!

PR #450, opened 2026-08-04
upsert

mcoughlin:
This PR introduces an upsert for async sessions.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -201,7 +201,35 @@ async_session_factory = None
 async_plain_session_factory = None
 
 
-class _AsyncVerifiedSession(SAAsyncSession):
+class _AsyncUpsertMixin:
+    """`upsert` shorthand for async sessions: update-or-insert by a natural key."""
+
+    async def upsert(self, model, *, by, values=None):
+        """Update the single ``model`` row matching ``by``, or insert a new one.
+
+        ``by`` maps attribute -> value identifying the row (typically a unique
+        key). If a row matches, each of ``values`` is assigned onto it; otherwise
+        ``model(**by, **values)`` is added. Both dicts may include relationship
+        attributes. Returns the instance and does not commit.
+        """
+        values = values or {}
+        instance = await self.scalar(
+            sa.select(model).where(*(getattr(model, k) == v for k, v in by.items()))
+        )
+        if instance is None:
+            instance = model(**{**by, **values})
+            self.add(instance)
+        else:
+            for key, value in values.items():
+                setattr(instance, key, value)
+        return instance
+
+
+class _AsyncPlainSession(_AsyncUpsertMixin, SAAsyncSession):
+    """Plain async session (no RLS check) carrying the `upsert` shorthand."""
+
+
+class _AsyncVerifiedSession(_AsyncUpsertMixin, SAAsyncSession):
     """Async counterpart of `_VerifiedSession`. Runs RLS verification on
     flush/commit using `async_bulk_verify`.
 
@@ -441,6 +469,7 @@ def init_db(
     )
     async_plain_session_factory = async_sessionmaker(
         bind=async_engine,
+        class_=_AsyncPlainSession,
         autoflush=autoflush,
         expire_on_commit=False,
     )


stefanv (commented):
Looks good; can we add a test?

mcoughlin:
@stefanv I am adding one that tests this to skyportal. Is that what you mean?

app/models.py
```diff
@@ -201,7 +201,35 @@ def bulk_verify(mode, collection, accessor):
 async_plain_session_factory = None
 
 
-class _AsyncVerifiedSession(SAAsyncSession):
+class _AsyncUpsertMixin:
+    """`upsert` shorthand for async sessions: update-or-insert by a natural key."""
+
+    async def upsert(self, model, *, by, values=None):
```
stefanv, line 207: Can we add type annotations here, or a NumPy-style docstring to show the types expected. E.g., what is `by`?

app/models.py
```diff
@@ -201,7 +201,35 @@ def bulk_verify(mode, collection, accessor):
...
+        """Update the single ``model`` row matching ``by``, or insert a new one.
+
+        ``by`` maps attribute -> value identifying the row (typically a unique
+        key). If a row matches, each of ``values`` is assigned onto it; otherwise
+        ``model(**by, **values)`` is added. Both dicts may include relationship
```
stefanv: What happens to models with implicit keys? Are those IDs ignored in the model construction, for the case when an insert is happening?

stefanv (commented):
I think it would be useful to have a test *inside baselayer*: use upsert twice, ensure that the second time the object is preserved but updated.

If we don't have non-end-to-end tests in baselayer, it's time :)

app/models.py
mcoughlin: added

mcoughlin: done.

app/models.py
mcoughlin: the new record looks like model(**{**by, **values}) so every key is present.

antoine-le-calloch:
@mcoughlin can you run a Claude /review on your side here pls? Mine found quite a lot of stuff.

mcoughlin:
@antoine-le-calloch mine found a few more things.

app/models.py
```diff
@@ -201,7 +201,72 @@ def bulk_verify(mode, collection, accessor):
...
+        ----------
+        model : `baselayer.app.models.DeclarativeMeta`
+            The mapped class of the target table.
+        by : dict of str to object
+            Attribute name to value, identifying at most one row: a natural key,
+            such as a unique column or a set of columns unique together. These
+            become attributes of the record on the insert path.
+        values : dict of str to object, optional
+            Attribute name to value, assigned to the row whether it was found or
+            created. Defaults to none, which makes the call a get-or-create.
```
stefanv:
```suggestion
            created. Defaults to None, which makes the call a get-or-create.
```

Claude, WTF!

stefanv: Also, what happens if values is None and the record isn't found? Then I guess you just make an empty record 🤷  Feels like `upsert` has very little use without `values` being required?

app/models.py
```diff
@@ -201,7 +201,72 @@ def bulk_verify(mode, collection, accessor):
...
+
+        The row is selected and then inserted, rather than through ``INSERT ...
+        ON CONFLICT``, so of two sessions that both miss, both insert and the
+        second to commit fails on the unique index. Callers racing for the same
+        key have to handle that.
+
+        A surrogate primary key the database fills in belongs in neither dict.
+        Values in ``by`` are passed to the constructor on the insert path, so
+        naming such a key writes it explicitly while its sequence stays where it
+        was, and the next insert to reach that value fails on the unique index.
```
stefanv: We can strip this gobbledygook.

test/conftest.py
```diff
@@ -0,0 +1,8 @@
+def pytest_addoption(parser):
+    # `load_env()` reads --config from sys.argv; declare it so pytest passes it through.
```
stefanv, line 2: Clever little 🤖

test/test_upsert.py
```diff
@@ -0,0 +1,146 @@
...
+        db["database"],
+    )
+
+
+@pytest.fixture
+def session_factory():
+    """A plain async session factory over a table created for the test."""
+
+    async def setup():
+        engine = create_async_engine(database_url())
```
stefanv, line 44: We'll refactor this general test infra later. Fine for now to have it here since we only have one piece of functionality tested.

stefanv:
Some minor things; will merge after next update.

app/models.py
mcoughlin: yes.

app/models.py
```diff
@@ -215,7 +215,98 @@ def bulk_verify(mode, collection, accessor):
...
+
+        Parameters
+        ----------
+        model : `baselayer.app.models.DeclarativeMeta`
+            The mapped class of the target table.
+        by : dict of str to object
+            Attribute name to value, identifying at most one row: a natural key,
+            such as a unique column or a set of columns unique together. These
+            become attributes of the record, and win over ``values`` where the
+            two name the same attribute.
```
stefanv: In the case where no record exists, these become part of the updated attributes.

Hrm, with conflicts, why make these win? Should probably check if they're equal and pass, or fail otherwise.

app/models.py
```diff
@@ -215,7 +215,98 @@ def bulk_verify(mode, collection, accessor):
...
+        values : dict of str to object, optional
+            Attribute name to value, assigned to the row whether it was found or
+            created. Defaults to None, which makes the call a get-or-create.
```
stefanv, line 195: What does None mean; create with only the primary key? That's a weird row to add.

app/models.py
```diff
@@ -215,7 +215,98 @@ def bulk_verify(mode, collection, accessor):
...
+        Notes
+        -----
+        ``values`` takes any mapped attribute; ``by`` is compared with ``==``, so
+        it takes columns and many-to-one relationships, and a collection raises
+        ``InvalidRequestError``. On a verified session the lookup runs through
+        ``model.select``, so a row the accessor cannot read is not found: the
+        call inserts instead and fails on the unique index rather than raising
+        ``AccessError``.
+
+        The row is selected and then inserted, not ``INSERT ... ON CONFLICT``,
```
stefanv: Is this true? Our selection mechanism means you need to identify a unique row through `by`, but we don't say that you need to specify a fully unique key. So, e.g., you may have a table that has:

```
id=1
apple=red

id=2
apple=green
```

If you use `apple=green`, you can select a unique row. If you then upsert twice on the non-unique apple field, one with `apple=green` and one with `apple=yellow`, won't it just overwrite? I don't think you'll trigger any non-unique constraints necessarily.

Anyway, maybe I'm off, just a thought.

app/models.py
antoine-le-calloch: Done !

app/models.py
antoine-le-calloch: Done !

app/models.py
antoine-le-calloch: Not the primary key, by. upsert(Group, by={"name": "x"}) creates Group(name="x"), which is the get-or-create case.

stefanv: I see now the intent; that the `by` fields become the ones that are modified. I still think this "which makes the call a get-or-create" statement is confusing. It simply means that, if the record does not exist, the `by` elements become the fields of the new record.

stefanv (approved):

PR #454, opened 2026-08-12
more auth backends

mcoughlin:
This PR makes more auth backends possible.

@camilledzt

### DIFF
--- a/doc/setup.md
+++ b/doc/setup.md
@@ -117,23 +117,33 @@ does need code, for example a bespoke federation endpoint, can subclass
 `social_core.backends.oauth.BaseOAuth2` anywhere importable and be named here by
 its dotted path; it does not have to live in baselayer.
 
-Providers are offered in the order listed, one login button each, and
-`debug_login` impersonates the first one so test mode exercises the same routes
-as production.
+The login page shows one button per provider, in the order listed. In [debug
+mode](#debug-mode), where sign-in succeeds without contacting any provider, the
+stand-in takes the name of the first provider you list, so development and
+production exercise the same login page and the same routes.
 
-### Identity across providers
+### Signing in with more than one provider
 
-An account is matched on `(provider, subject)` — the provider's own stable id
-for the user. `use_unique_user_id` (default `true`) is what makes the subject the
-key rather than the email address: emails get reassigned, and two providers
-reporting one address are not thereby the same person.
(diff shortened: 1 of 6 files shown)


mcoughlin:
@antoine-le-calloch tagging you so you see this here.

stefanv (commented):
Looks OK to me but needs docs.

antoine-le-calloch (approved):
LGTM

doc/setup.md
```diff
@@ -79,7 +79,61 @@ See [below](#configuration) for more information on modifying the baselayer conf
...
+          oidc_endpoint: https://iam.example.org
+```
+
+Anything under `settings` is exported as `SOCIAL_AUTH_<NAME>_<KEY>`, so a
+provider speaking standard OpenID Connect needs no code at all. A provider that
+does need code, for example a bespoke federation endpoint, can subclass
+`social_core.backends.oauth.BaseOAuth2` anywhere importable and be named here by
+its dotted path; it does not have to live in baselayer.
+
+Providers are offered in the order listed, one login button each, and
```
stefanv: This is a bit murky; how is debug_login supposed to be used?

doc/setup.md
```diff
@@ -79,7 +79,61 @@ See [below](#configuration) for more information on modifying the baselayer conf
...
+`debug_login` impersonates the first one so test mode exercises the same routes
+as production.
+
+### Identity across providers
+
+An account is matched on `(provider, subject)` — the provider's own stable id
```
stefanv: How do we handle people logging in via multiple services? E.g., same user, but from Google and GitHub?

doc/setup.md
```diff
@@ -79,7 +79,61 @@ See [below](#configuration) for more information on modifying the baselayer conf
...
+for the user. `use_unique_user_id` (default `true`) is what makes the subject the
+key rather than the email address: emails get reassigned, and two providers
+reporting one address are not thereby the same person.
+
+That matters as soon as a second provider is enabled. An app that wants a
```
stefanv: I think this is too technical for user docs: explain the interface and usage, not the technical details underlying.

stefanv (commented):
Please improve the docs, and then merge when you consider ready.

doc/setup.md
mcoughlin: Requires the same email, noting this in there.

PR #456, opened 2026-08-17
Remove comment in fill_conf_values recipe

antoine-le-calloch:
- Remove the two comment lines added above the `find` call in `fill_conf_values`
- They are recipe lines, so make echoes them on every run


mcoughlin (approved):
LGTM

PR #457, opened 2026-08-17
Add opt-in quiet mode to status(), replaying captured output on failure

mcoughlin:
This PR adds opt-in quiet mode to status(), replaying captured output on failure (ported from the Fritz version).


mcoughlin:
@thomasculino @antoine-le-calloch

antoine-le-calloch (approved):

PR #458, opened 2026-08-18
On finish guard

mcoughlin:
This PR stops a dead connection's rollback from raising out of on_finish (run into sometimes on Fritz).

@thomasculino @antoine-le-calloch


antoine-le-calloch (approved):
LGTM

PR #459, opened 2026-08-19
get emails from github login

mcoughlin:
This PR adds custom code to verify emails when using Github login.


antoine-le-calloch (approved):
LGTM

PR #460, opened 2026-08-22
quiet client-caused 4xx in log_exception

mcoughlin:
This PR fixes some 4xx that made it to Sentry as "errors."

### DIFF
--- a/app/handlers/base.py
+++ b/app/handlers/base.py
@@ -12,7 +12,7 @@ from json.decoder import JSONDecodeError
 import sqlalchemy
 import tornado.escape
 from tornado.log import app_log
-from tornado.web import RequestHandler
+from tornado.web import HTTPError, RequestHandler
 
 from ...log import make_log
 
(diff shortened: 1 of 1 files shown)


app/handlers/base.py
```diff
@@ -144,7 +145,14 @@ def log_exception(self, typ=None, value=None, tb=None):
             "Unauthorized",
         ]
         v_str = str(value)
-        if any(exception in v_str for exception in expected_exceptions):
+        # A 4xx is the caller's fault; scanner traffic would otherwise be
```
stefanv: Maybe something compact like:

```suggestion
        # HTTP 4xx is for client error. Do not log these.
```

app/handlers/base.py
```diff
@@ -136,6 +136,7 @@ def write_error(self, status_code, exc_info=None):
         self.render("loginerror.html", app=cfg["app"], error_message=str(err))
 
     def log_exception(self, typ=None, value=None, tb=None):
+        # Expected conditions raised as plain exceptions, with no status to test.
```
stefanv: I don't understand this comment.

stefanv (commented):
Looks good overall; would be great if we can clarify the comments here.

app/handlers/base.py
mcoughlin: fixed!

PR #461, opened 2026-08-26
allow apps to provide a tools/clear_data.py that will run on db_clear.

antoine-le-calloch:
`db_clear` drops the database but leaves behind the on-disk data tied to it. The folders holding that data are app-specific, so baselayer cannot know about them. This adds an optional hook: if the app provides `tools/clear_data.py`, `db_clear` runs it before dropping the database. Apps without that file are unaffected, and if the script exits non-zero (for instance because the user declined a confirmation prompt). Keeping the hook here, rather than overriding `db_clear` in each app's Makefile, avoids having several entry points for the same command. Used by skyportal/skyportal#6603.


antoine-le-calloch:
I'm not happy with this solution, let's move everything to SkyPortal instead. This doesn't make a lot of sense and is dangerous in the case of an already existing file with that name.

PR #462, opened 2026-08-26
Improve makefile help

antoine-le-calloch:
Move targets defined in several makefiles to their own overrides section, instead of duplicating them with different descriptions. This is the simplest solution I have found to avoid having two different descriptions for the same command (on `make help`), rather than going through #461, which I am not happy with. I still think the overall idea behind a more generic `db_clear` was not bad.

`make help` example with my latest SkyPortal PR, skyportal/skyportal#6603, which overrides `make db_clear`: 

image


stefanv:
Thanks, Antoine. Happy to merge this, but also curious: shouldn't we just let later Makefiles override earlier ones? Ultimately, there is only one command that will be executed.

antoine-le-calloch:
> Thanks, Antoine. Happy to merge this, but also curious: shouldn't we just let later Makefiles override earlier ones? Ultimately, there is only one command that will be executed.

Hum not sure to understand what do you mean.

stefanv:
We process the help strings for Makefiles, starting with baselayer, then moving on to others. If a command in baselayer is also defined in one of the others, simply use the latest definition in the help string.

antoine-le-calloch:
> We process the help strings for Makefiles, starting with baselayer, then moving on to others. If a command in baselayer is also defined in one of the others, simply use the latest definition in the help string.

@stefanv Oh yeah, okay. That was my first implementation, but as a developer, I found it useful to keep the information that baselayer also has a version. But yeah, for the user, it’s better to just hide it. I’ll move back to my old version.

PR #463, opened 2026-08-28
Return failed sign-ins to the login page instead of a 500

mcoughlin:
This PR returns failed sign-ins to the login page instead of a 500.


antoine-le-calloch (approved):
LGTM

PR #464, opened 2026-09-04
Fix bulk verify composite pk

antoine-le-calloch:
- Fix a 500 on any commit whose session holds a row from a join table declared
  with `join_model(..., composite_pk=True)`. `bulk_verify` and
  `async_bulk_verify` built their query from `record_cls.id`, but those tables
  have no surrogate `id`: `join_model` sets the attribute to `None`, so
  `columns=[record_cls.id]` became `columns=[None]`, SQLAlchemy emitted
  `SELECT NULL AS anon_1`, and the next line's `accessible_row_ids_sq.c.id`
  raised `AttributeError: id`.
- Derive the primary key from `primary_key_keys()` in both functions, then
  select, join and filter on every one of its columns. `ComposedAccessControl`
  already did this for the same reason; the two verification functions had been
  missed.
- Keep the plain `IN (...)` form when the primary key is a single column, so the
  statement generated for ordinary models is identical to before. The row-wise
  `(a, b) IN ((...), ...)` form is used only for composite keys, where index
  behaviour differs and this check runs on every commit.
- Read the primary keys from the identity map instead of from `record.id`, which
  the async function already did to avoid a synchronous lazy load raising
  `MissingGreenlet` on an expired object. The sync function now shares that path.


app/models.py
```diff
@@ -305,6 +289,40 @@ def primary_key_keys(cls):
...
 
 
+def inaccessible_pks_stmt(collection, accessible_rows, pk_cols):
+    """Select the primary keys of ``collection`` that are absent from
+    ``accessible_rows``.
+
+    Keyed on ``pk_cols`` rather than a surrogate ``id`` so join tables built with
+    ``composite_pk=True`` are handled. PKs come from the identity map because
+    ``record.id`` can sync-lazy-load an expired object and raise MissingGreenlet
+    under async.
```
stefanv: Future developers won't understand what this means; explain without the context of the missing ID why we do it this way.

stefanv: This function is called `stmt` — does it produce a statement or? No type annotation, and no NumPy-style docstring, so hard to know what it produces.

app/models.py
```diff
@@ -162,29 +162,22 @@ def bulk_verify(mode, collection, accessor):
...
-        collection_ids = {record.id for record in collection}
+        pk_cols = [getattr(record_cls, key) for key in primary_key_keys(record_cls)]
 
         # vectorized query for ids of rows in the session that
         # are accessible
-        accessible_row_ids_sq = record_cls.query_records_accessible_by(
-            accessor, mode=mode, columns=[record_cls.id]
-        ).subquery()
+        accessible_rows = record_cls.query_records_accessible_by(
+            accessor, mode=mode, columns=pk_cols
```
stefanv, line 172: I suspect this is the real fix. Can we add a comment maybe? E.g., checking to see which rows are accessible by the user; these rows are accessed by primary key, and primary key consists of various columns.

stefanv (approved):
The fix looks correct to me, @antoine-le-calloch. Have you tested this? If so, I think we can merge. (I asked for some trivial docs/comments updates.)

stefanv:
Also, acknowledging that this is a long-outstanding bug; a bit surprised we managed to get away with surrogate keys for so long.

antoine-le-calloch:
@stefanv yep, I have tested it. I agree, and to be honest, I was also surprised to discover that this had been happening all this time. I found this bug because after fixing another bug in SkyPortal, my other PR https://github.com/skyportal/skyportal/pull/6697 was no longer passing CI tests because of this one. So one bug fix led me to another, ahah.

stefanv:
Thank you, @antoine-le-calloch.

PR #466, opened 2026-09-11
Improve logging

antoine-le-calloch:
- Only log that the database is not migrated on the first try and then every 30s, instead of on every retry.
- Drop the "could not connect to migration manager" line, it repeated the same information.
- Log the number of refreshed tables instead of listing them all.
- Print the attempt number when the database is not reachable while creating the tables.
- Serialise log writes in `watch_logs` so lines from different services no longer cut into each other.
- Keep tailing a log file that is rotated or removed instead of staying stuck on a dead file handle.
- Back off the log discovery interval up to 60s instead of jumping straight to it.
- Silence the raw supervisor output on `make stop`.
- Clean up dead code and redundant variables.


mcoughlin (approved):
LGTM

PR #467, opened 2026-09-11
Clean up change from #449

antoine-le-calloch:
- Drop `verify_permissions()` and `verify_and_commit()`, no longer called by any handler; `self.Session()` checks on commit
- Drop the `AccessError` branch in `write_error()`, it tested the exception class and never fired
- Make `resolve_pooler()` public, it is imported from outside the module
- Rename `EXECUTEMANY_PAGESIZE` to `INSERTMANYVALUES_PAGE_SIZE`, after the parameter it sets
- Share one set of engine arguments between the sync and async engines
- Factor the row collection and grouping out of the sync and async verify paths
- Remove the unused `_get_client()` dask helper
- Simplify the auth lookups, the connection retries and `create_tables()`, and trim the comments left over from #449


doc/dev.md
```diff
@@ -147,48 +147,33 @@ can be replaced with `DBSession()` with no arguments.
...
-codebase: `async_session_factory()` verifies on commit, and
-`async_plain_session_factory()` skips that check. `app/access.py` uses the plain one:
-
-```
-from baselayer.app import models
-
-async with models.async_plain_session_factory() as session:
-    ...
-```
+`DBSession()` is the exception. It is scoped on `session_context_id`, a context variable
```
stefanv: Is there ever a reason to call DBSession directly? What is that use-case?

doc/dev.md
```diff
@@ -147,48 +147,33 @@ can be replaced with `DBSession()` with no arguments.
...
+that `BaseHandler.prepare()` sets to a fresh value on every request, so inside a request
+every call returns that request's one session. Outside a request the variable keeps its
+default of `None` and every call in the process returns a single shared session: in a
+microservice, a script, or work handed to a thread, one caller's `rollback()` then
+discards another caller's pending work. Use `new_session()` there.
 
-Reach the async factories through the module rather than importing their names directly:
-`init_db()` rebinds those globals, so `from baselayer.app.models import
-async_plain_session_factory` captures the `None` they hold before that call.
+The async sessions are opt-in per handler. The sync engine and session stay authoritative
```
stefanv, line 178: An example, with code, like you had but a bit more explicit, would still be quite nice. Could of course be a future update to the docs.

stefanv (commented):
Cleanups like these are very welcome. Thanks for improving the docs.

stefanv (approved):

antoine-le-calloch:
> Cleanups like these are very welcome. Thanks for improving the docs.

Done!

PR #469, opened 2026-09-13
Fix the flaky server readiness check

antoine-le-calloch:
`verify_server_availability` probed nginx while the app workers were still starting. nginx banned the refused upstreams for `server.fail_timeout` (20s) and the tests ran against the status server's 503.

Wait for the internal app ports to answer before probing nginx.


tools/test_frontend.py
```diff
@@ -46,7 +46,16 @@ def all_services_running():
     return running if return_code in (0, 3) else False
 
 
-def verify_server_availability(url, timeout=180):
+def app_workers_listening(cfg):
+    for i in range(cfg["server.processes"]):
```
stefanv, line 50: Why do you need to cycle through these? Shouldn't one passing be enough?

antoine-le-calloch: Each worker is a separate nginx upstream with its own failure counter, and the
`api` pool excludes worker 0. One 200 only proves worker 0 is up, while nginx
just banned the others for fail_timeout, so API requests get a 503 from the
backup status server.

tools/test_frontend.py
```diff
@@ -46,7 +46,16 @@ def all_services_running():
...
+        try:
+            requests.get(f"http://localhost:{cfg['ports.app_internal'] + i}", timeout=1)
```
stefanv: Instead of hitting the index/login page, and having the DB initialize (all quite heavy), we could streamline this a bit by adding a very light JSON endpoint, something like `/baselayer/health` or `/baselayer/up`. We can then probe that directly. Asked Claude what that could look like:

```
  class HealthHandler(tornado.web.RequestHandler):
      """Liveness probe: answers as soon as the IOLoop is accepting requests.

      A 200 means this worker process is serving, not that the DB is up etc.
      """

      def get(self):
          self.set_header("Content-Type", "application/json")
          self.set_header("Cache-Control", "no-store")
          self.write({"status": "success", "data": {"pid": os.getpid()}})

      head = get
```

tools/test_frontend.py
```diff
@@ -46,7 +46,16 @@ def all_services_running():
...
+        except requests.exceptions.ConnectionError:
```
stefanv:
```suggestion
        except requests.exceptions.RequestException:
```

Also covers ReadTimeout (TCP handshake, but then doesn't get us back our data). 

Then:

```
          if r.status_code != 200:
              return False
```

PR #470, opened 2026-09-20
Only run the dev-only services when the config asks for them

antoine-le-calloch:
`fake_oauth2` runs in every deployment, and nginx proxies `/fakeoauth2` to it, so `/fakeoauth2/auth` answers even when `server.auth.debug_login` is false and the fake backend is unused. It redirects to any `redirect_uri` it is given.

Two changes:

- The service now reads `server.auth.debug_login` at startup and idles when it is false, so the port is not served. This check belongs in the service because the supervisor config is rendered when the Docker image is built, where the value can differ from the config mounted at run time.
- Its supervisor config becomes a template, so the program is not set up at all when `debug_login` is false, the same way `pgbouncer` and `dask` already work. `rspack` gets the same treatment with `env.debug`: it already exits on its own in production, and it no longer shows up as an exited program in `supervisorctl status`.


services/fake_oauth2/fake_oauth2.py
```diff
@@ -29,6 +31,13 @@ def post(self):
 
 
 env, cfg = load_env()
+log = make_log("fake_oauth2")
+
+if not cfg["server.auth.debug_login"]:
+    log("server.auth.debug_login is false: not serving the fake OAuth2 endpoints")
+    # Idle rather than exit so supervisor doesn't restart-loop.
+    while True:
```
stefanv, line 39: Should be able to configure supervisord to see a different exit code and read meaning into it. But if not immediately obvious how, don't worry about it.

stefanv (approved):
LGTM; go ahead and merge when ready.

services/fake_oauth2/fake_oauth2.py
antoine-le-calloch: Good idea. Done !

stefanv:
Thanks, looks good to me!

